PR #2, opened 2020-10-16
migrate code from original djinni repo & add some basic integration testing

### LINKED ISSUE #1: Add integration-testing for the generator
The current generator implementation has no tests. There is just global integration-testing, that involves building a working program with the generated gluecode.
Once the generator is extracted from the uber-repo, it will need its own testing, to allow at least some QA. 
I want to write some basic integration-tests, that make sure that input A produced the expected output B. This should be enough as a starting point.

jothepro:
fixes #1 

This is quite a big commit, so I will list what I changed and you will basically have to trust me 😕 

- I copied all sourcefiles from the [djinni repo](https://github.com/cross-language-cpp/djinni/tree/master/src) over to the src folder of the template.
- I added some integration testing and integrated it into github actions. This is a first step towards at least regression-testing the generator in isolation.


jothepro:
the pipeline will not run in this MR, as it adds github-actions for the first time. Have a look at the green pipeline here: https://github.com/jothepro/djinni-generator-1/actions

a4z:
awesome! we do it as with the other, just take the change as a start and if something is required to add/change/fix we will iterate over it

a4z (approved):

PR #8, opened 2020-11-15
Add version information

### LINKED ISSUE #7: Add --version parameter
As a user I'd like to be able to run `djinni --version` to find out the currently installed version. This makes it easier to verify what version has been installed on a system.

One idea on how to achieve this is that on release creation the name of the version tag is somehow injected into the build action.

jothepro:
fixes #7 

djinni now supports a `--version` parameter that returns the version of the version of the cli! 🥳 
The exact version number is automatically generated from the latest git tag by this awesome sbt plugin: https://github.com/dwijnand/sbt-dynver

example:
```
> djinni --version
djinni generator version 0.1.0
```

The version information is also prepended to the help text:
```
> djinni --help
djinni generator version 0.1.0
Usage: djinni [options]

  --help                   
  --version                
[...]
```

I'll point out the interesting parts in the PR in some comments, as it came out pretty bloated for such a small change.


jothepro:
btw sorry for messing up the git history. :P

project/plugins.sbt
```diff
@@ -1 +1,2 @@
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
\ No newline at end of file
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
+addSbtPlugin("com.dwijnand" % "sbt-dynver" % "4.1.1")
```
jothepro, line 2: This plugin does all the magic!

src/main/scala/djinni/Main.scala
```diff
@@ -97,7 +98,9 @@ object Main {
       }
 
       override def showUsageOnError = false
+      head("djinni generator version", Main.getClass.getPackage.getImplementationVersion)
       help("help")
+      version("version")
```
jothepro, line 103: the scopt library already automatically supports the `--version` parameter. All you need is to define a `head()` with the version information.

jothepro:
I guess because I messed up the git history, the github actions won't run. I'm too lazy to fix it if you don't force me to. 🙄 
Here is the successful pipeline for reference: https://github.com/jothepro/djinni-generator-1/runs/1402590527

src/it/scala/djinni/MetadataIntegrationTest.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import org.scalatest.GivenWhenThen
+import org.scalatest.Matchers.{convertToAnyShouldWrapper, startWith, include}
+
+class MetadataIntegrationTest extends IntegrationTest with GivenWhenThen {
+  it("should output version information") {
+    When("calling `djinni --version`")
+    val versionDescription = djinni("--version")
+    Then(s"the version description should be returned")
+    versionDescription should startWith ("djinni generator version")
```
jothepro, line 11: This is not an amazing test. All I do is test if the cli output starts with the correct string. I do not verify that a properly formatted version is appended.

The test at least ensures the program won't crash without outputting anything.
One could in the future add a regex evaluation of the version to get even more confidence.

a4z: this test is fine, I can add an additional test in the conan recipe that I have, here I currently also just test that djinni returns something, but I can adopt that to call --version and parse the output, once the next version is tagged

a4z (approved):
awesome!

Issue #20, opened 2021-01-05
option --objc-header-out seems to be missing


a4z:
As a user, I would like to be able to define where all the public interface headers of as library using djinni are

--

The output location for Objective C header can not be set like those for C++ and JNI header.

For cpp and jni header and source files, there are those options that play together

*  --cpp-out <out-folder>
* --cpp-header-out <out-folder>
* --cpp-include-prefix

<br>

*  --jni-out
*  --jni-header-out <out-folder>
*  --jni-include-prefix <prefix>
*  --jni-include-cpp-prefix

There is an option for `--objc-include-prefix` , but no one for `--objc-header-out`

Since the objective C header are similar a part of the public interface like the other headers, it would be nice being able to tell the generator where to put the objective C header files to.

PR #21, opened 2021-01-06
Objc header out

a4z:
suggestion to solve #20


jothepro:
NICE! Please add the new option to the documentation at `docs/cli-usage.md`

jothepro (approved):
I'm feeling a little bit proud that all of our investment in understanding how djinni works and updating it's infrastructure finally led to this change that adds true value for all djinni users. Thank you, Harald! 💪

Issue #14, opened 2020-12-04
the #include directives in the generated code do no use the folder structor of the support lib


a4z:
currently, generated code has, for example in case of jni bindings, this include statement

`#include "djinni_support.hpp"`

but I think it should be

`#include <djinni/jni/djinni_support.hpp>`

so it is possible to not having to add various include directives, or just use standard include if the support lib is installed into the file system / root system 

alternative opinions are welcome !

jothepro:
This seems to make sense. I wonder if we should introduce this as a breaking change or if it should be configurable as a parameter, so existing build setups won't fail.

a4z:
I think this is a very good question, one that might pop up from time to time .... 

one side of me says: add an option
an other side says: no options, it will become too many execution paths and more code to maintain, complexity will add up

let us thing a moment about that, and maybe collect more information

- 
for the very subject, I am not sure that we break too much, because, basically what we would do is saying: please use default include path (to include) , the rest is picked up automatically , and instead of 3 you have just 1 include path to manage.

paulo-coutinho:
Hi,

I think the same thing. It need be a more specific path for djinni, because sometimes it has conflicts with other headers.

But it need be applied to all support lib.

And in my opinion you can use quotes:

```
#include "djinni/jni/djinni_support.hpp"
```

Because it is not in the compiler/system path, but custom/app path.

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

Thanks.

a4z:
The compiler will find it anyway, its just a question of the search order,
If the support lib comes with a conan or hunter package, what it sooner or later will, or ist installed what it could be, its not a " path first.
But in general I do not have any strong preferences about <or", once we will have that implemented, doing the one or the other will be easy.

paulo-coutinho:
Sure.

With quotes will be compatible with generated code by djinni:

![image](https://user-images.githubusercontent.com/395096/104057218-87e65a00-51d0-11eb-87fb-7c809e592652.png)

I think (again, in my opinion) that we can make it more similar.

What need for this? Can i make a PR for this?

a4z:
Sure Paulo! 
Contributions are very welcome! 

Please see https://djinni.xlcpp.dev/contributing/

a4z:
I just found out , this is already possible,
 
the `--jni-base-lib-include-prefix` can do that.

`djinni .....   --jni-base-lib-include-prefix djinni/jni/ ....` 
will generate
[...]

PR #22, opened 2021-01-18
Generate full path to support lib include files

a4z:
This targets #14 ,
generated files now include the full path to files includede from
the support lib.

This shall simplify the include setup
Incldue support-lib install_prefix/include is now enough.

This change should not break any existing setups that inlcude full
paths.
However, to restore the previous behavious,
set `--jni-base-lib-include-prefix` to an emppty string


a4z:
If there are no different opinions that this is a bad idea, I will start adoption the tests so they reflect the new reality

a4z:
Did some additional testing, I think it is safe to apply the change.

We will need to have to add an additional documentation issue for the next release where this changes is described, 
but since this will contain more than just this pr, I would like having that in an extra PR

jothepro (changes requested):
Please document the default values in `--help` + the [documentation](https://djinni.xlcpp.dev/djinni-generator/cli-usage/)

a4z:
I was actually planning to do the docs in the context of #23

jothepro:
hm. I'm not sure if were talking about the same kind of documentation.
Documenting the default value in the `--help` command requires a code change [here](https://github.com/a4z/djinni-generator/blob/include-prefix/src/main/scala/djinni/Main.scala#L162), I see this change in the scope of this PR. What I mean: 
```
> djinni --help
...
--jni-include-prefix <prefix>   The prefix for #includes of JNI header files from JNI C++ files (default: djinni/jni/).
```

This is how default values are documented for all other properties.

The docs contain just a copy of the --help output, so [here](https://github.com/a4z/djinni-generator/blob/include-prefix/docs/cli-usage.md#jni) the same change is required.

a4z:
ah, ok, awesome catch!!

its those section that have the doc for this change
```
      opt[String]("jni-base-lib-include-prefix").valueName("...").foreach(x => jniBaseLibIncludePrefix = x)
        .text("The JNI base library's include path, relative to the JNI C++ classes.")
```
and
```
      opt[String]("objc-base-lib-include-prefix").valueName("...").foreach(x => objcBaseLibIncludePrefix = x)
        .text("The Objective-C++ base library's include path, relative to the Objective-C++ classes.")
```

any suggestion for the text?

What about
`The djinni-support-lib JNI include path. Deprecated, only available for backward compatibility.`
and
`The djinni-support-lib Objective-C include path. Deprecated, only available for backward compatibility.`
for both ?

jothepro:
I don't understand the purpose of this parameter. :/

jothepro:
Silly me, it's the params that you have set default values for 🙈 .
I don't see them as deprecated, it's just that they have a useful default value already set. 🤔

a4z:
they are deprecated, except for code that is generated with old generator version, where you need to set it to ""

jothepro:
What if someone decided to not install the base lib as it is, but copy it's files to a different folder structure? 🤔 
What I want to say: to me it seems like these parameters allow you to configure something that normally you do not want to configure. But in my opinion that does not make them deprecated, it just makes them not very interesting. The fact that you have decided to set a new default value for them is a proof to me that they are not useless :P
Deprecating them would mean that we discourage users from using these parameters because they may be removed in the future. I don't think that this is the case, or is it?

a4z:
you describe what I say with `except in serious exotic setups I do not want to think about.` :-) 
No plans for removal of that option.

So simply update the text to 
`The djinni-support-lib Objective-C include path.`
and add some hint on the website docs that you do not want to use it ?

jothepro:
What I'd put there: 

    The JNI base library's include path (default: djinni/jni/).

I prefer this because it follows the same pattern as the other parameters documentation.
I think it is important to have the default value documented. I'd prefer not to have a link or any deeper explanation in there because it would clutter the cli output.

But I think I'm starting to see why you want to add a "deprecated" or "do not use" hint. It's not user friendly to have the configuration parameters list bloated with options that basically nobody needs to set... I'm not sure what to think about this right now. This may apply to a lot of options.

a4z:
sounds good, thanks!
Will fix it tomorrow

jothepro (approved):
Awesome!

Issue #11, opened 2020-11-17
enums, records and possible others, should not be generated in uppercase but as written in the IDL file


a4z:
As a user, I want that my enums, records and other entities are generated as specified in the djinni idl file, and not be converted into uppercase letters.

Origin: 
https://github.com/cross-language-cpp/djinni-intellij-plugin/issues/18

a4z:
This name generation can  be controlled via the `--ident-java-enum` command line option, like all (most) other naming can be controlled via such options.

For file names that is for sure a good idea, for something like enum values, I do not know ...

There seems to be no option that says, do not convert that text, it goes through some conversion. 
And that feels like unwanted feature creep.

So fixing/improving that would be, finding an option to let` text in - text out` happen

Until we fix that, please use the `--ident-java-enum` and see if any of ``FooBar`, `fooBar`, `foo_bar`, `FOO_BAR`, `m_fooBar`` gives you better results than you have now.

a4z:
The `foo_bar` style is the style that does not convert anything. So if you want to avoid conversions, use this option in the corresponding `--ident-*` arguments.

We could change that, and I know how, and I thing we should, since the default should give users the least surprise, and not the pre configured coding stile from the original authors.

However, this would be a breaking change also, so we need to think a little bit more about that.

Either starting to version the idl files, or having an additional flag that selects the default behaviour ... 
Feedback is welcome!

jothepro:
So if one chooses the `foo_bar` option, there will be no conversion of identifiers, no matter if they are TrainCase, camelCase, ...? That's weird. 
I'd prefer improving the documentation on this instead of changing anything that may break someones code.

jothepro:
Setting `foo_bar` as default option sounds like a reasonable workaround. 👍

a4z:
We can actually break the default behaviour if we provide the `--whatever-option value` that restore the previous behaviour.
So users will have an easy way to not break existing generated code by copy and past a string. 

I think this is the way to go, but we should test it with #14 first
So I will implement it there, since this has zero impact for users.
And based on that we can communicate this kind of flag change.

Then we can announce the upcoming flag change and see if we get any feedback.
That way is not as fast as I would like to have it, but I do not want to annoy potential or actual users.

a4z:
Solved via #24 and #28, all settings are configurable by a user and documented now

PR #24, opened 2021-01-20
Update docs

a4z:
Add cli doc for recent change , #14 / #22
Check that docs for #20 / #21 is in place
Improve Identifyer Style section , for #11

### DIFF
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -150,7 +150,6 @@ Identifier styles (ex: `FooBar`, `fooBar`, `foo_bar`, `FOO_BAR`, `m_fooBar`)
 | `--ident-objc-type-param ...` |
 | `--ident-objc-local ...` |
 | `--ident-objc-file ...` |
-|   |
 
 The _Identifier Style_ is a text conversion hint on how to generate names from the djinni idl file.
 
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -191,5 +191,5 @@ The best way to find out how that works is playing around with different values 
 
 !!! note
 
-    Some defaults have been set while back, that, for example enum names are transformed to uppercase letters.
-    At the moment we do not want to change that behavior, but please be aware that we think the default should always be no text transformation for all settings.
+    Some identifiers do have text transformation enabled by default. 
+    Nevertheless make sure to **explicitly set** text transformation styles if you need them! All text transformation may be disabled by default in a future release of the generator.
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -187,7 +187,7 @@ This way you can adopt code generation to some extend to your existing coding st
 
 The best way to find out how that works is playing around with different values for different identifier options.
 
-If you wish not text transformation to happen, use the `foo_bar` style.
+**If you wish not text transformation to happen, use the `foo_bar` style.**
 
 !!! note
 
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -80,7 +80,7 @@ djinni \
 | `--jni-include-prefix <prefix>` | The prefix for #includes of JNI header files from JNI C++ files. |
 | `--jni-include-cpp-prefix <prefix>` | The prefix for #includes of the main header files from JNI C++ files. |
 | `--jni-namespace ...` | The namespace name to use for generated JNI C++ classes. |
-| `--jni-base-lib-include-prefix ...` | The JNI base library's include path, relative to the JNI C++ classes. |
+| `--jni-base-lib-include-prefix ...` | The JNI base support library's include path (default: djinni/jni/). |
 
 ### Objective-C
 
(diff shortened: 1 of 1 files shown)


a4z:
I wonder if there is a way to preview the docs local, especially when playing around with mkdocs specials, like the 
`!!! note` section

richey-v:
> I wonder if there is a way to preview the docs local, especially when playing around with mkdocs specials, like the
> `!!! note` section

I've used [Python mkdocs](https://www.mkdocs.org/) locally before on a project.  That project was hosted on a local Gitlab instance (not github) so I spent some time ensuring my MarkDown was compatible and rendered as intended with both the Gitlab and mkdocs rendering.

jothepro:
you can preview the docs by running `mkdocs serve`.
This will render the docs with the default mkdocs theme.

The admonition blocks (`!!! note`) require a [special plugin](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) to be activated.

I opened [an issue](https://github.com/cross-language-cpp/cross-language-cpp.github.io/issues/11) to improve the preview-ability in each repository. Thanks for your feedback, @a4z

docs/cli-usage.md
```diff
@@ -150,6 +150,46 @@ Identifier styles (ex: `FooBar`, `fooBar`, `foo_bar`, `FOO_BAR`, `m_fooBar`)
...
+
+
+The best way to find out how that works is playing around with different values for different identifier options.
+
+**If you wish not text transformation to happen, use the `foo_bar` style.**
+
+!!! note
+
+    Some defaults have been set while back, that, for example enum names are transformed to uppercase letters.
+    At the moment we do not want to change that behavior, but please be aware that we think the default should always be no text transformation for all settings.
```
jothepro, line 195: I have trouble understanding what the first sentence wants to say. Additionally I think we should give a reason for why we recommend to disable text transformation.

I propose a note like this:

```suggestion
    Some types do have text transformation enabled by default. To avoid breaking existing builds, this will not be changed in the future.
    Nevertheless it is recommended to disable all text transformation to avoid issues in the target language.
```

Am I right with the reason for our recommendation? I did not follow the discussion on this in detail.

docs/cli-usage.md
```diff
@@ -150,6 +150,46 @@ Identifier styles (ex: `FooBar`, `fooBar`, `foo_bar`, `FOO_BAR`, `m_fooBar`)
 | `--ident-objc-type-param ...` |
 | `--ident-objc-local ...` |
 | `--ident-objc-file ...` |
+|   |
```
jothepro, line 153: You are talking about default values below. Could you document them here and in the `--help` output, so I can see at a glance what the default is on each type? This would be really helpful!

Why did you add the new blank table row at the end?

docs/cli-usage.md
```diff
@@ -150,6 +150,46 @@ Identifier styles (ex: `FooBar`, `fooBar`, `foo_bar`, `FOO_BAR`, `m_fooBar`)
...
+
+```
+public enum SomeEnum {
+    VALUE,
+    ;
+}
+```
+As you see, _VALUE_ is now in upper case letters.
+
+If you use `--ident-java-enum foo_bar` then _Value_ will not be transformed into upper case and the following Java code will be generated.
```
jothepro, line 176: This is so weird 🙈  
If changing the Identifier names would not potentially break so many builds I'd prefer more speaking names for them:

- `camelCase`
- `TrainCase`
- `snake_case`
- `none` <== This is what `foo_bar` currently is.
- ...

jothepro: What do you think about renaming them? This just came to my mind a few days ago. I don't see it in the scope of this PR.

jothepro (commented):
Looks great so far! But I have a few optional questions and ideas for improvement.

docs/cli-usage.md
a4z: might still be a breaking change. 
we could add `none` as a keyword to start a smooth transition process ....

docs/cli-usage.md
a4z: this is possible, but I need to refactor_read this from the source to get the info out
its actually not difficult, but I might take some time

docs/cli-usage.md
a4z: its not just types that have transformations, but also enum values, member fields .....

yes, we can improve the second sentence,  I would like to transport the message that in future that might change
so please start setting everything explicit now, or so ... just to open a door

a4z: I will think a little later about how that could be improved, if you have further suggestion, I am happy to merge them in

docs/cli-usage.md
jothepro: I'm fine with opening a new issue for this if it requires some more effort.

docs/cli-usage.md
jothepro: just want to talk with you about this on a theoretical basis. :P I'm aware that this is a major breaking change.

docs/cli-usage.md
jothepro:
```suggestion
    Some identifiers do have text transformation enabled by default. 
    Nevertheless make sure to **explicitly set** text transformation styles if you need them! All text transformation may be disabled by default in a future release of the generator.
```

docs/cli-usage.md
jothepro:
```suggestion

```

jothepro: I created #25 for the documentation of the default values.

jothepro (approved):
Approved given that the empty trailing row in the ident-style parameter table is removed. 👍

docs/cli-usage.md
a4z: ah, yes, this empty line,  I added it there to create an additional line at the end, but my VS code preview looks of course different than the mkdoc one

PR #27, opened 2021-01-21
add asdf install instructions

jothepro:
Additionally I added a nice badge that shows the total downloads of artifacts.


a4z (commented):
Thanks!
I did not know there is a dl counter, that's great!

a4z (approved):

PR #29, opened 2021-01-31
update sbt to 1.4

joprice:
This updates sbt to 1.4, which enables use of `sbtn`, the experimental
thing client to sbt server also built with graal. This allows entering
the sbt shell locally very quickly, since you only wait once for the
initial build to load. You can also quickly run commands like `sbtn
test` without fearing that it will take an eternity to start. This makes
sbt feel more like bazel or cmake.


jothepro (approved):
LGTM!

a4z (approved):

PR #32, opened 2021-03-09
Use named parameters for better readability

freitass:
Added named parameters for the ident styles to help identifying what goes where. I also formatted them similarly to how they were defined (3 lines each), preserving the original position of each param.


jothepro (approved):
I like that! Thank you :)

PR #35, opened 2021-03-19
Integration tests for PR 34

jothepro:
added integration tests that would have detected the bug that was fixed in https://github.com/cross-language-cpp/djinni-generator/pull/34

### DIFF
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -1,6 +1,7 @@
 package djinni
 
 import org.scalatest.GivenWhenThen
+import org.scalatest.Matchers.{convertToAnyShouldWrapper, equal}
 import org.scalatest.prop.TableDrivenPropertyChecks._
 
 class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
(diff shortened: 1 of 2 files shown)


src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -78,5 +79,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
         assertFileContentEquals(idlFile, OBJCPP, objcppFilenames)
       }
     }
+
+    it("should be able to only generate Java output") {
+      val outputPath = "src/it/resources/result/only_java_out"
+      When("calling the generator with just `--java-out`")
+      val output: String = djinni(s"--idl src/it/resources/all_datatypes.djinni --java-out $outputPath")
```
jothepro, line 86: Without the fix, this line throws an exception because the called command returned a non-zero value (1). The tests would fail because of that.

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -78,5 +79,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+      Then("the generator should successfully generate just java output")
+      output should equal ("Parsing...\nResolving...\nGenerating...\n")
+      assertFileExists(s"$outputPath/AllDatatypes.java")
```
a4z, line 89: Is it possible to add some assert that file not exist for Objective-C?
just to be sure that this is not generated

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -78,5 +79,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+    }
+
+    it("should be able to only generate Obj-C output") {
+      val outputPath = "src/it/resources/result/only_objc_out"
+      When("calling the generator with just `--objc-out`")
+      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --objc-out $outputPath")
+      Then("the generator should successfully generate just objc output")
+      output should equal ("Parsing...\nResolving...\nGenerating...\n")
+      assertFileExists(s"$outputPath/AllDatatypes.h")
+      assertFileExists(s"$outputPath/AllDatatypes.mm")
```
a4z, line 99: Is it possible to add some assert that file not exist for Java?
just to be sure that this is not generated

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -78,5 +79,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+    }
+
+    it("should be able to only generate C++ output") {
+      val outputPath = "src/it/resources/result/only_cpp_out"
+      When("calling the generator with just `--cpp-out`")
+      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --cpp-out $outputPath")
+      Then("the generator should successfully generate just cpp output")
+      output should equal ("Parsing...\nResolving...\nGenerating...\n")
+      assertFileExists(s"$outputPath/all_datatypes.hpp")
```
a4z, line 108: Is it possible to add some assert that file not exist for Objective-C and Java? 
just to be sure that this is not generated

src/it/scala/djinni/GeneratorIntegrationTest.scala
jothepro: mh. good idea 👍

jothepro: hm, now that I think about it, this test would not make a lot of sense:

If no `--java-out` parameter is given, then the generator does not know where to put the java files, even if he erroneously decided to create them. As a tester I don't know where to expect these files, so I cannot test for their absense. 

If I wanted to be sure that the generator did not generate Java-files in the objc-out directory, I can already rely on the [tests above](https://github.com/cross-language-cpp/djinni-generator/blob/main/src/it/scala/djinni/GeneratorIntegrationTest.scala#L59). They make sure that all files are written to where they should be.

jothepro:
@a4z do you think I should create a release when this is fixed? The whole issue seems quite problematic to me. I didn't spot it earlier because I still have not migrated my toolchain to the `cross-language-cpp` djinni 🤦 . I think my build would be broken without the fix in #34

src/it/scala/djinni/GeneratorIntegrationTest.scala
a4z: agreed, 
 `find ./ -name "*.java"` , and with h/mm for the java case,  in the current working directory and output path, might give no 100% security nothing has been written (without creating a chroot environment) but it might be a nice placebo to have.
Question is if it's too much hassle to create .... and it's also not super important
So I approve and let the decision to you :-)

a4z:
yes, please feel. free to tag a version, docu update and bug fix are a good reason

a4z (approved):

PR #38, opened 2021-03-29
Set Swift bridging header location

### LINKED ISSUE #31: swift bridging header output location
Currently the header is generated into `--objc-out` , but it should go into`--objc-header-out` , or a user defined location

The naming is also something that could also be checked, if you choose `foo-umbrella` for the header name (note, no extension)  `fooumbrellaVersionNumber` is generated. 
Not optimal, but the `-umbrella` part seems to be what some people use.

We also need to check if the code that is generated is fit for 2021 and laster 

And when looking over that, maybe other things will be found.
So ideas and suggestions are welcome, if possible they will be integrated.

a4z:
Fixes #31 by setting the Swift bridge header location to the Objective-C
output header location.


src/main/scala/djinni/Main.scala
```diff
@@ -308,7 +308,7 @@ object Main {
       None
     }
     val objcSwiftBridgingHeaderWriter = if (objcSwiftBridgingHeaderName.isDefined && objcOutFolder.isDefined) {
-      val objcSwiftBridgingHeaderFile = new File(objcOutFolder.get.getPath, objcSwiftBridgingHeaderName.get + ".h")
+      val objcSwiftBridgingHeaderFile = new File(objcHeaderOutFolder.get.getPath, objcSwiftBridgingHeaderName.get + ".h")
```
jothepro, line 311: dumb question from me: Wouldn't it be even cooler if one could not only specify the name of the bridging header but the full path? 🤔

a4z: This is an absolute valid question. 

There is a/some relation between the bridging / umbrella header , and the public Objective-C included files.
The bridging header must find them, and as it is now, the exiting implementation will do so wit this solution.

I see also no value to put them into a different place, since this will only make everything more complex for no, to me visible, benefit.

jothepro (approved):
I haven't tested it but it looks reasonable to me. I'm looking forward to using the generated bridging headers!
A few final thoughts:
- This looks like a breaking change to me? If we do semantic versioning this could be the reason to bump the generator version to 1.0.0?
- I think this should be documented in the docs/cli-help. If you don't do so now, can you pls create an issue so we don't forget it?
- I think it would be cool to have a test for this. It should be pretty straight forward. Maybe I will contribute a test in the next days :)

a4z:
I do not think that the current change will break anything, it is a header and it will be found , since it is there where the other headers that needs to be found are. So that's fine.

Yes, of course should that be in docs, change log management is still something on the todo list (pls a list of resolved issues and PRs since the last release)

jothepro:
I'm not sure if we are talking about the same thing regarding documentation. What I meant was that it should be documented here in the docs https://djinni.xlcpp.dev/djinni-generator/cli-usage/#objective-c and in the output of `djinni --help` where the header file is created (in `--objc-header-out`)

a4z:
Ah, thanks Johan, good catch!
You should un approve until I have updated the docs ...  so I do not accidental merge this PR

jothepro (changes requested):
update documentation

jothepro (approved):

PR #42, opened 2021-04-09
bugfix: remove trailing ";" from toCpp function in JNI generator

### LINKED ISSUE #41: syntax error in generated C++ code for JNI
the generated C++ code for JNI contains stray ";" characters in places where the C++ syntax does not allow them.

the fix is trivial, we will upload a pull request to fix this in a minute

Neusoft-Technology-Solutions:
that's a syntax error in C++,
triggering the Clang/GNU warning "-Wextra-semi"

### DIFF
--- a/src/main/scala/djinni/JNIGenerator.scala
+++ b/src/main/scala/djinni/JNIGenerator.scala
@@ -218,7 +218,7 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
         w.wl
         if (spec.cppNnType.nonEmpty) {
           def nnCheck(expr: String): String = spec.cppNnCheckExpression.fold(expr)(check => s"$check($expr)")
-          w.w("static CppType toCpp(JNIEnv* jniEnv, JniType j)").bracedSemi {
+          w.w("static CppType toCpp(JNIEnv* jniEnv, JniType j)").braced {
             w.wl(s"""DJINNI_ASSERT_MSG(j, jniEnv, "$jniSelf::fromCpp requires a non-null Java object");""")
             w.wl(s"""return ${nnCheck(s"::djinni::JniClass<$jniSelf>::get()._fromJava(jniEnv, j)")};""")
           }


jothepro:
Can you provide me an example input configuration and generation output on when this happens, so I can reproduce it? I'm a bit confused by the fact that this is a change in the generators output but does not create an error in our integration tests. It seems like we have a blind spot in the tests...

Neusoft-Technology-Solutions:
We used Djinni generator 0.3.1.

**Example:**

example.djinni:
```djinni
MyInterface = interface +j +o {
   someFunction(someValue: i32): bool;
}
```

run with
```bash
djinni --idl example.djinni --jni-out=. --cpp-nn-type=abc
```

Important is the ```--cpp-nn-type=...``` option.

the resulting C++ code is:

```c++
static CppType toCpp(JNIEnv* jniEnv, JniType j) {
   DJINNI_ASSERT_MSG(j, jniEnv, "MyInterface::fromCpp requires a non-null Java object");
   return ::djinni::JniClass<MyInterface>::get()._fromJava(jniEnv, j);
};
```
Note the last ";" after the function.

a4z:
djinni assert message is not a function but a macro , defined like that

```
#define DJINNI_ASSERT_MSG(check, env, message) \
    do { \
        ::djinni::jniExceptionCheck(env); \
        const bool check__res = bool(check); \
        ::djinni::jniExceptionCheck(env); \
        if (!check__res) { \
            ::djinni::jniThrowAssertionError(env, __FILE__, __LINE__, message); \
        } \
    } while(false)
```

and I do not see the double `;` if used like shown ... it should expand to valid code.

please give us some time to evaluate that

Neusoft-Technology-Solutions:
We are not talking about DJINNI_ASSERT_MSG here.
We are talking about the ";" in the last line, following the "}"

a4z:
ah, ok, now I see it , thanks !

jothepro (approved):
LGTM, I'm tempted to not add tests to cover this edge case right now.
Thanks for supplying the fix, @Neusoft-Technology-Solutions !!

Neusoft-Technology-Solutions:
thank you!

a4z (approved):
awesome, thank you!

PR #43, opened 2021-04-13
Add Python support

freitass:
Bringing Python support from original Dropbox-managed repository almost as-is and added some integration tests for it.


a4z:
Since this is required for https://github.com/cross-language-cpp/djinni-support-lib/pull/24, I would say we pull it in and prepare a new release. I have some success on my local mach with that, for both,  Linux and Mac.

I think these PRs should be done in a way leading us to 1.0.0, 
maybe a 0.99.0 , and once everything is fine we change both , generator and support lib , to 1.0.0 .. ?

jothepro:
We haven't talked about a versioning concept in djinni yet but I think we should stick to [semantic versioning](https://semver.org/). I would be fine with making an exception to send a signal that "we are feature complete now", but personally I'd like to wait with that until C# support is also in master.

jothepro:
First of all: Super cool, @freitass! Thanks for your contribution!!!

I am missing documentation on about how the djinni integration works:

- "[Interface Definition Language (IDL)](https://djinni.xlcpp.dev/djinni-generator/idl/)":
    - Updated example listings to include `+p` modifier
    - Datatypes: How do C++ types map to Python types (date, map, ...)?
    - Exceptions: How does the Python integration handle Exceptions?
- "[Using the generator](https://djinni.xlcpp.dev/djinni-generator/cli-usage/)": Documentation of existing command line parameters (basically the same information that is output by `djinni --help`.
- "[Use generated Code](https://djinni.xlcpp.dev/djinni-generator/generated-code-usage/)": 
    - What files are generated?
    - A bit of background how the Python <-> C++ interaction works. I am aware that this is new for us all so I think it cannot be super in-depth, but a few sentences would help me and hopefully others to understand what is going on.

You can find the Markdown-files that generate the documentation in `docs`.

src/main/scala/djinni/Main.scala
```diff
@@ -215,34 +223,56 @@ object Main {
...
+      opt[File]("pycffi-out").valueName("<out-folder>").foreach(x => pycffiOutFolder = Some(x))
+        .text("The output folder for PyCFFI files (Generator disabled if unspecified).")
+      opt[String]("pycffi-package-name").valueName("...").foreach(x => pycffiPackageName= x)
+        .text("The package name to use for the generated PyCFFI classes.")
+      opt[String]("pycffi-dynamic-lib-list").valueName("...").foreach(x => pycffiDynamicLibList= x)
+        .text("The names of the dynamic libraries to be linked with PyCFFI.")
+      opt[File]("c-wrapper-out").valueName("<out-folder>").foreach(x => cWrapperOutFolder = Some(x))
+        .text("The output folder for Wrapper C files (Generator disabled if unspecified).")
+      opt[String]("py-import-prefix").valueName("<import-prefix>").foreach(pyImportPrefix = _)
+        .text("The import prefix used within python genereated files (default: \"\")")
```
jothepro, line 237:
```suggestion
        .text("The import prefix used within python generated files (default: \"\")")
```

jothepro (commented):
Wow... I am migrating the C# support in parallel and I have a feeling like the python integration is so much more complex... 😮 At least it generates a lot more gluecode 🙈 .
At a first glance the only missing feature in the generator seems to be the ability to set a header import prefix for the support-lib (equivalent to `--jni-base-lib-include-prefix`. Do you plan to add it?

src/it/resources/expected/my_flags/python/my_flags.py
```diff
@@ -0,0 +1,17 @@
...
+
+from enum import IntEnum, unique
+
+@unique
+class MyFlags(IntEnum):
+    Flag1 = 0
+    Flag2 = 1
+    Flag3 = 2
+    No_Flags = 3
+    All_Flags = 4
```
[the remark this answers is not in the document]
freitass: It should be investigated.

freitass:
> At a first glance the only missing feature in the generator seems to be the ability to set a header import prefix for the support-lib (equivalent to `--jni-base-lib-include-prefix`. Do you plan to add it?

We don't have any new features planned for Djinni, but it'd be cool to have that option to make it similar to the other languages!

freitass:
> I am migrating the C# support in parallel and I have a feeling like the python integration is so much more complex... 😮

Indeed! It'll be much harder to document Python in https://djinni.xlcpp.dev/djinni-generator/generated-code-usage/. I believe that's our only big piece of documentation missing right now.

src/it/resources/expected/my_flags/python/my_flags.py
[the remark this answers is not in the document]
jothepro: Maybe because `IntFlag` was only introduced in Python 3.6 ([link](https://docs.python.org/3/library/enum.html#enum.auto)), which was released in late 2016. Do you have a minimum python3 version in mind, that should be supported?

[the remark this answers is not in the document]
a4z: 3.6 should be fine, earlier versions are not supported anymore, and we need to think to the future, not the past. 
3.6 will end this year, with supporting that version  we would be as much backward compatible as it makes sense.

paulo-coutinho:
Hi.

First, good job man, very thanks for that. Im reading the files changes to review and have two questions:

1. The cwrapper output files can be work with other things like flutter or is specific for python?
2. The yaml files can be used with python generator too?

Thanks.

src/it/resources/expected/my_flags/python/my_flags.py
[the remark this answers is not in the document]
freitass: I just pushed a change to use `IntFlag` for Djinni flags in Python. Resulting code looks like this: https://github.com/freitass/djinni-generator/blob/93c9e4cbb952ae17558123d5b5b2515a2a152da1/src/it/resources/expected/my_flags/python/my_flags.py#L9-L16

docs/generated-code-usage.md
```diff
@@ -65,3 +65,66 @@ Generated files for Objective-C / C++ are as follows (assuming prefix is `DB`):
...
+
+Compile the Python extension module (CFFI) by executing `pycffi_lib_build.py` while providing all C
+Wrapper header files (`.h`) as arguments. The resulting shared library will enable Python to access
+your C++ library through the CFFI bridge.
+
+### Known limitations of the generator
+
+* External types defined in YAML are not yet supported.
+* Use of non-nullable pointers is not yet supported.
+* There are no arguments for configurable header prefixes (or use of existing arguments).  All includes are unqualified.
```
jothepro, line 128: I've introduced

- `--c-wrapper-include-prefix <prefix>`
- `--c-wrapper-include-cpp-prefix <prefix>`
- `--c-wrapper-base-lib-include-prefix <prefix>`

So from my point of view header prefixes are as configurable as they can get. Am I missing something?

src/it/resources/expected/my_enum/python/my_enum.py
```diff
@@ -6,9 +6,8 @@
 
 from djinni import exception # this forces run of __init__.py which gives cpp option to call back into py to create exception
 
-from enum import IntEnum, unique
+from enum import IntEnum
 
-@unique
```
jothepro: Is there a reason why the `@unique` annotation had to be removed? Is it incompatible with `IntFlag`?

docs/generated-code-usage.md
freitass: You're right. I didn't realize how many of the limitations we've already addressed so far!

src/it/resources/expected/my_enum/python/my_enum.py
freitass: I concluded [from the docs](https://docs.python.org/3/library/enum.html) it only applies to `Enum` and since the code is generated anyways I didn't see the value of having it (even if it was applicable).

freitass:
@jothepro I believe we've improved all points mentioned in [this comment](https://github.com/cross-language-cpp/djinni-generator/pull/43#issuecomment-818498894), can you confirm?

jothepro (approved):
Awesome!!! Let's bring this to master! 🚀  Thank you for your passionate contribution, @freitass!

I did a final review and these are the last two things that caught my eyes:

- Do you know wether record [extension](https://github.com/freitass/djinni-generator/blob/python/docs/idl.md#extensions) is supported in Python? It's not mentioned in the docs.
- I found some more autogenerated headers that were still missing a "was". I've made a [PR for this](https://github.com/freitass/djinni-generator/pull/3).

I have not yet tested the generated code on my own! My review is based on wether the generators output looks reasonable to me. I trust you that you have successfully used the Python integration at least in a simple test setup once!
I am ok with the risk that I do not at all understand the details of this implementation. We'll learn about those details when things break 😉 .

a4z (approved):
Guys, you are doing an awesome job! 

Thanks a lot!!

a4z:
and I agree, let's bring that to master, we then can enable uint tests for the support lib and start iterating over the states of both.
maybe we will also get additional testers when its there to use ....

jothepro:
👍 ok, @freitass I'll wait for you to merge my PR and answer my question. Once I got your response I will merge into master and make a release. Then we can start focusing on the support-lib.

a4z:
what we would need would be something like a latest generator build, to download, because now the support lib still doesn't have any way to get it (accept, getting the code and building it at every PR, what is not optimal) 
@jothepro , do you see an easy way to always have a latest to download after a merge into master?

jothepro:
I don't see a problem there (yet). I think it is a good design decision to make the support-lib depend on a specified release version instead of an undefined 'latest' state of the generator.
No, I do not have a decent idea for that right now.

a4z:
this means, we need a release now to run the unit tests .... 
and if we (would) need one or 2 fast iterations for fixing things in the generator/support lib , this would mean one or 2 fast releases for nothing ... I think  I need to think about that 
.... latests would not be undefined, it's the latest .... maybe attaching it as artefact to a gh action , we will see

freitass:
> * Do you know wether record [extension](https://github.com/freitass/djinni-generator/blob/python/docs/idl.md#extensions) is supported in Python? It's not mentioned in the docs.
> * I found some more autogenerated headers that were still missing a "was". I've made a [PR for this](https://github.com/freitass/djinni-generator/pull/3).

Check and check! ✅

PR #47, opened 2021-04-23
Remove any hardcoded references to std::experimental

freitass:
[As mentioned in the Python PR on the support lib](https://github.com/cross-language-cpp/djinni-support-lib/pull/24#discussion_r617139250), there were a few places in the generator where explicit references to `std::experimental` had been left behind. The options presented at that point to solve it were to either hardcode `std::optional`/`std::nullopt` everywhere or use the command-line parameter for the optional type to replace the aforementioned references.

This PR represents a third option, which is to replace the three instances of `std::[experimental::]nullopt` with code that does not specify any type:

- Evaluate the optional type directly in a boolean context (`false` means `!= std::[experimental::]nullopt`)
- Default-construct an `std::[experimental::]optional` object with `{}` (which is equivalent to `std::[experimental::]nullopt`)

This might be an acceptable solution until we decide (or find time) to deal with the `std::[experimental::]optional` situation as a whole.

### DIFF
--- a/src/it/resources/expected/all_datatypes/cwrapper/dh__all_datatypes.cpp
+++ b/src/it/resources/expected/all_datatypes/cwrapper/dh__all_datatypes.cpp
@@ -170,7 +170,7 @@ djinni::Handle<DjinniRecordHandle> DjinniAllDatatypes::fromCpp(const ::AllDataty
 }
 
 djinni::Handle<DjinniOptionalRecordHandle> DjinniAllDatatypes::fromCpp(std::optional<::AllDatatypes> dc) {
-    if (dc == std::experimental::nullopt) {
+    if (!dc) {
         return nullptr;
     }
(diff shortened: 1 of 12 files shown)


a4z (approved):

a4z:
Code change is looks correct, thanks Leandro!
I just wonder, the update of the generator to generate the expected code, that happens in an other PR ?

jothepro (approved):

jothepro:
@a4z I think you missed the change in `src/main/scala/djinni/CWrapperGenerator.scala`. That's where the output of the generator has been adjusted.

a4z:
> @a4z I think you missed the change in `src/main/scala/djinni/CWrapperGenerator.scala`. That's where the output of the generator has been adjusted.

True, somehow I saw it, but at the same time not ... thanks for pointing me to that!

PR #49, opened 2021-04-30
add new `--jni-generate-main` parameter

jothepro:
~~I realized that a huge portion of the docs regarding the yaml files was not yet migrated to the new docs. So I did this and added some other small things while doing so.~~

This PR has been repurposed to introduce a new parameter, sorry for the chaos 🙈 . I have created 2 new PRs #50 , #51 to reduce the complexity of this PR.

This is still a Draft because ~~I need to add a concept for C++ <-> C++/CLI custom types.~~
~~@freitass maybe you can help out with that? 🙈~~ the new parameter needs further discussion.


docs/cli-usage.md
```diff
@@ -43,7 +43,7 @@ djinni \
 | `--help` | Print help  |
 | `--version` | Print version |
 | `--idl <in-file>` | The IDL file with the type definitions, typically with extension `.djinni`. |
-| `--idl-include-path <path> ...` | An include path to search for Djinni `@import` directives. Can specify multiple paths. |
+| `--idl-include-path <path> ...` | An include path to search for Djinni `@import` and `@extern` directives. Can specify multiple paths. |
```
jothepro, line 46: As you can see [here](https://github.com/cross-language-cpp/djinni-generator/blob/main/src/main/scala/djinni/parser.scala#L46), the include path seems to affect both `@import` and `@extern`.

a4z: good catch!

a4z (approved):

src/main/scala/djinni/JNIGenerator.scala
```diff
@@ -35,6 +35,15 @@ class JNIGenerator(spec: Spec) extends Generator(spec) {
...
 
+  if(spec.jniGenerateMain) {
+    createFile(spec.jniOutFolder.get, "djinni_jni_main." + spec.cppExt, w => {
+      w.wl("// AUTOGENERATED FILE - DO NOT MODIFY!")
+      w.wl("// The file included here provides default JNI_OnLoad and JNI_OnUnload implementations.")
+      w.wl("// Disable the generation of this file with `--jni-generate-main=false`.")
+      w.wl(s"#include " + q(spec.jniBaseLibIncludePrefix + "djinni_jni_main.hpp"))
+    })
+  }
+
```
jothepro, line 46: This is where the magic happens. The file `djinni_jni_main.cpp` will be written to `--jni-out`. It contains nothing but an include of the header `#include djinni/jni/dijnni_jni_main.hpp` that implements `JNI_OnLoad` & `JNI_OnUnload`. This way, if the user builds a shared library from the generated sources, it will automatically have the correct ABI for use with JNI. 

This does not help if the user builds a static lib from the generated sources & links it from another shared library. In that case the user would have to include the header from the support-lib himself again. 

The good thing: It doesn't matter if this is part of a static libraries ABI. It will get stripped away when linked by another library anyways.

jothepro (commented):
I am abusing this PR for changes that are required to support the latest changes in https://github.com/cross-language-cpp/djinni-support-lib/pull/26.

For this I added a new commandline parameter `--jni-generate-main`.

The paramater, if `true`, generates a file `djinni_jni_main.cpp` that includes the `djinni/jni/djinni_jni_main.hpp` header provided by djinni-support-lib. That way by default the user automatically gets a default JNI_OnLoad implementation if he builds a shared library from all sources.

This is a breaking change because it generates a new type of file. This requires the upcoming djinni-support-lib v1.0.0, that comes with the required header file, to work properly.

This PR is starting to get huge and messy... I can try to split it into pieces if this helps you. But you'll have to ask me to do that extra work ;)

src/main/scala/djinni/JNIGenerator.scala
a4z: I am not convinced that is needed, why not always let the user include `dijnni_jni_main.hpp` , if we choose to go that route .... in that way the usage is always the same, no matter it the support lib is used static or dynamic.
Or, just disable a dynamic support lib ... 

let us think more about the question, but I really would like to avoid additional command line arguments that we can/do not test.

docs/modularization.md
```diff
@@ -0,0 +1,162 @@
...
+    static ::djinni::LocalRef<jobject> JniType fromCpp(CppType c) { return /* your magic here */; }
+
+    using Boxed = Record1;
+};
+```
+
+```
+// For C++ <-> C++/CLI
+// TODO
+```
```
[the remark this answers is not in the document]
freitass: I'll spend some time on it this weekend (hopefully tonight)! 👌

src/main/scala/djinni/JNIGenerator.scala
jothepro: I am trying to solve the following: 

If the user would have to include the `djinni_jni_main.hpp` himself (which would be the easiest solution, I agree), the instructions would sound something like this:

"If building for JNI, include `djinni_jni_main.hpp` from the support lib somewhere in your target sources."

The question is: where should the header be included? In a normal djinni project setup all non-generated C++ code should be platform independent. So the user would have to

- add source file that includes `djinni_jni_main.hpp`, which only is added to the build sources if building for JNI or
- add a preprocessor condition that only includes the header if building for JNI.

This adds complexity to the users build configuration.

If we would decide to go a different direction and tell the user to include a sourcefile provided by the support lib that implements `JNI_OnLoad`, like you suggested in our slack discussion, the install instructions would be:

"If building for JNI, add `djinni_jni_main.cpp` to your target sources"

This comes with the same overhead. The user has to add logic to his build configuration that includes the file if he wants to build for JNI, otherwise not.

The new option `--jni-generate-main` moves that required complexity away from the user. Now he doesn't even have to know about the internals and the generated code will just work without him adding additional logic for it.

a4z: we have 4 different scenarios
* static - our init
* static -custom init
* dynamic - our init
* dynamic - custom init
shall we put this into a own PR, where we might play all that 4 rules, and see how it looks like in real when doing so,
or shall we do it here?

jothepro: To reduce the complexity of this PR I plan to create separate smaller PRs for the stuff from this PR that we agree on and that doesn't need further discussion (documentation, improved tests). 
That way this PR will shrink in size once we merged the smaller changes. Does that make sense or am I missing sth?
In the end this PR should more or less only contain stuff regarding the newly introduced parameter. So let's discuss the new parameter here. 👍 

I'm not sure if I understand you 4 scenarios correctly:
1. **static support-lib** linked into shared library with **default init** *(-> when `--jni-generate-main=true`)*
2. **static support-lib** linked into shared library with **custom init** *(-> `--jni-generate-main=false` + custom implementation. Requires the user to write platform specific configuration logic.)*
3. **shared support-lib with default init** linked by shared library with **default init**
4. **shared support-lib with default init** linked by shared library with **custom init**
5. ~~dynamic support-lib with custom init linked by shared library~~ *(-> If building the support lib as dynamic library, it is self contained and only needs to initialize itself. Therefore no custom implementation is required.)*

a4z: sounds good!
yes, 
I think 3 and 4 are highly unlikely use cases anyway, so we can/should focus on 1 and 2 
The default behaviour should be, I do nothing, have no custom init, and it just works. 
Is that possible and if not, what is the minimum action a user has to do?

jothepro: I tried to sketch the different variations.

1 & 3 require no knowledge of the internals if the proposed `--jni-generate-main` is introduced:

djinni_jni (4)

a4z: ok, now I think I understand the whole process, please let me know if I summaries that now correctly:

* djinni_jni_main.cpp get generated as all the other files from the IDL, 
* it will therefore be included in the list of generated files, 
* user will get it into users_shared.lib by just adding all generated files, as today
* the default setting is to generatedjinni_jni_main.cpp, so everything just works out of the box! 

If I got it now right, it makes a lot of sense.
Awesome as always, Johannes!
Thanks for your patience explaining me the new details!!

jothepro: Yes, that's correct!

freitass (approved):

jothepro:
Although you both approved this PR I am not 100% sure that I have your approval to merge this because this PR got quite chaotic. Is further discussion needed? Pls give me a quick thumbsup if you think that the current state of this branch can be merged. I've tested the changes against the PR in the djinni-support-lib, everything should be working.

freitass:
It looks good! :D

a4z:
merge, and if anyone thinks there is chaos, anyone can fix it later ;-)

Issue #48, opened 2021-04-24
Do not require all supported languages to be specified in YAML files


freitass:
Djinni generator currently fails to parse a YAML file unless all of the supported languages are specified; there's an unhandled exception `java.util.NoSuchElementException: key not found: <key>` for any missing field (as can be seen in [this run](https://github.com/cross-language-cpp/djinni-support-lib/runs/2424293948)). This issue becomes more relevant as more languages get supported by the generator.

paulo-coutinho:
Hi,

After upgrade to version 0.5.0 and generate my files i get error:

```
![image](https://user-images.githubusercontent.com/395096/116300362-30ff9300-a775-11eb-850f-1ba2ed8160b2.png)
```

```
Exception in thread "main" java.util.NoSuchElementException: key not found: translator
	at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:107)
	at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:105)
	at djinni.YamlGenerator$.metaFromYaml(YamlGenerator.scala:227)
	at djinni.resolver.package$.$anonfun$resolve$1(resolver.scala:52)
	at djinni.resolver.package$.$anonfun$resolve$1$adapted(resolver.scala:38)
	at scala.collection.LinearSeqOptimized.foreach(LinearSeqOptimized.scala:75)
	at scala.collection.LinearSeqOptimized.foreach$(LinearSeqOptimized.scala:72)
	at scala.collection.mutable.MutableList.foreach(MutableList.scala:33)
	at djinni.resolver.package$.resolve(resolver.scala:38)
	at djinni.Main$.main(Main.scala:363)
	at djinni.Main.main(Main.scala)
[...]

PR #52, opened 2021-05-03
Do not require YAML definitions for a language that is not being generated

freitass:
When resolving YAML, if missing the definitions for one of the languages that are being generated a descriptive error message will be printed. For instance, when running djinni with `--cppcli-out`, if any of the keys for `cs` are missing the generation will fail with the following error:

```
/path/to/file.yaml (1.1): missing 'cs' definitions
```

That was achieved by making most fields of `MExtern` optional, requiring a call to `get` to be able to access their values; the check that generates the error message above ensures that any `Option[...]` field will have `Some(value)` (i.e. not `None`) by the time we call `get` on it.

In order to have the command line parameters available when resolving the YAML definitions I had to shuffle some code in `Main.scala`, causing a few lines that were originally in the "generation" phase to be executed during the "resolving" phase.

Fixes https://github.com/cross-language-cpp/djinni-generator/issues/48.

### DIFF
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -224,8 +224,11 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
     it("should be able to parse yaml files without all languages defined") {
       val outputPath = "src/it/resources/result/only_yaml_out"
       When("calling the generator with ")
+      a [RuntimeException] should be thrownBy {
+        djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
+      }
       noException should be thrownBy {
-        djinni(s"--idl src/it/resources/date_no_cs.djinni --yaml-out $outputPath")
+        djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
       }
     }
   }
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -224,12 +224,9 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
     it("should be able to parse yaml files without all languages defined") {
       val outputPath = "src/it/resources/result/only_yaml_out"
       When("calling the generator with ")
-      a [RuntimeException] should be thrownBy {
-        djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
-      }
-      noException should be thrownBy {
-        djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
-      }
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
+      noException should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
     }
   }
 }
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -234,7 +234,9 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
 
     it("should gracefully fail when a required language definition is incomplete") {
       val outputPath = "src/it/resources/result/only_yaml_out"
-      When("calling the generator with ")
+      Given("an IDL file that depends on a YAML type definition that misses a required key in the `objc` definition")
+      When("calling the generator")
+      Then("the generation should fail gracefully if code for Objective-C is generated")
       a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
     }
   }
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -225,8 +225,13 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
       val outputPath = "src/it/resources/result/only_yaml_out"
       When("calling the generator with ")
       a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
-      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
       noException should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
+    }
+
+    it("should gracefully fail when a required language definition is incomplete") {
+      val outputPath = "src/it/resources/result/only_yaml_out"
+      When("calling the generator with ")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
     }
   }
 }
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -223,8 +223,12 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
 
     it("should be able to parse yaml files without all languages defined") {
       val outputPath = "src/it/resources/result/only_yaml_out"
-      When("calling the generator with ")
+      Given("an IDL that depends on a YAML type definition that misses the `cs` key")
+      When("calling the generator  with `--cppcli-out` to generate C# gluecode")
+      Then("the generation should fail gracefully")
       a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
+      When("calling the generator with `--java-out` to generate just Java gluecode")
+      Then("the generator should not fail, because the `cs` type definition is not needed")
       noException should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
     }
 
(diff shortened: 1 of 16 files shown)


src/main/scala/djinni/CppCliGenerator.scala
```diff
@@ -158,7 +158,7 @@ class CppCliGenerator(spec: Spec) extends Generator(spec) {
         case p: MPrimitive => "."
         case MOptional => "."
         case MDate => "."
-        case e: MExtern => if (e.cs.reference) "->" else "."
+        case e: MExtern => if (e.cs.reference.get) "->" else "."
```
freitass, line 161: Every call to `get` will find `Some(value)` (i.e. not `None`).

src/main/scala/djinni/Main.scala
```diff
@@ -460,8 +451,16 @@ object Main {
...
 
-
+    // Resolve names in IDL file, check types.
     try {
+      resolver.resolve(meta.defaults, idl, outSpec) match {
+        case Some(err) =>
+          System.err.println(err)
+          System.exit(1); return
+        case _ =>
+      }
```
freitass, line 461: This block of code was moved down so we could use the information from `outSpec`. (I'm not 100% sure about this one.)

src/main/scala/djinni/meta.scala
```diff
@@ -37,41 +37,41 @@ object MExtern {
...
+    typename: Option[String],
+    boxed: Option[String], // Java typename used if boxing is required, must be an object.
+    reference: Option[Boolean], // True if the unboxed type is an object reference and qualifies for any kind of "nonnull" annotation in Java. Only used for "record" types.
+    generic: Option[Boolean], // Set to false to exclude type arguments from the Java class. This is should be true by default. Useful if template arguments are only used in C++.
+    hash: Option[String], // A well-formed expression to get the hash value. Must be a format string with a single "%s" placeholder. Only used for "record" types types with "eq" deriving when needed.
     writeToParcel: String, // A well-formed expression to write value into android.os.Parcel. Must be a format string with a single "%s" placeholder. Only used for "record" types types
     readFromParcel: String // A well-formed expression to read value from android.os.Parcel. Must be a format string with a single "%s" placeholder. Only used for "record" types types
```
freitass, line 62: Even if Java code is being generated, these two fields are not required as they have default values specified.

src/main/scala/djinni/YamlGenerator.scala
```diff
@@ -192,46 +193,54 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
...
-      nested(td, "java")("hash").toString,
-      if (nested(td, "java") contains "writeToParcel") nested(td, "java")("writeToParcel").toString else "%s.writeToParcel(out, flags)",
-      if (nested(td, "java") contains "readFromParcel") nested(td, "java")("readFromParcel").toString else "new %s(in)"),
+      nested(td, spec.javaOutFolder.isDefined, "java", "typename", _.toString),
+      nested(td, spec.javaOutFolder.isDefined, "java", "boxed", _.toString),
+      nested(td, spec.javaOutFolder.isDefined, "java", "reference", _.asInstanceOf[Boolean]),
+      nested(td, spec.javaOutFolder.isDefined, "java", "generic", _.asInstanceOf[Boolean]),
+      nested(td, spec.javaOutFolder.isDefined, "java", "hash", _.toString),
+      nested(td, false, "java", "writeToParcel", _.toString) getOrElse "%s.writeToParcel(out, flags)",
+      nested(td, false, "java", "readFromParcel", _.toString) getOrElse "new %s(in)"),
```
freitass, line 221: Fields not required. Here's where we set their default values when not specified.

freitass:
Support lib tests pass with this code (C# wasn't tested).

src/main/scala/djinni/Main.scala
jothepro: Only solution I see for this is to pass all `<lang>outFolder` parameters to `resolve` one by one?

```
resolver.resolve(meta.defaults, idl, javaOutRequired: javaOutFolder.isDefined, objcOutRequired: objcOutFolder.isDefined, objcppOutRequired: objcppOutFolder.isDefined, cppcliOutRequired: cppcliOutFolder.isDefined, cppOutRequired: cppOutFolder.isDefined)
```

One advantage I could see in this approach is that the `metaFromYaml` internals wouldn't be as verbose, because the `.isDefined` check only happens once in the function call:

```
MExtern.Cpp(
      nested(td, "cpp")("typename").toString,
      nested(td, "cpp")("header").toString,
      nested(td, "cpp")("byValue").asInstanceOf[Boolean]),
      nested(td, cppOut, "cpp", "typename", _.toString),
      nested(td, isRequired: cppOutRequired, "cpp", "header", _.toString),
      nested(td, isRequired: cppOutRequired, "cpp", "byValue", _.asInstanceOf[Boolean])),
```

Plus the code could be a bit easier to understand because the input for `resolve()` is defined more precise.

I'm sure you have thought about that for a longer time than me though. Did you consider this? The annoying part about it is that all parameters have to be passed to resolved & then passed to `metaFromYaml`.

src/main/scala/djinni/YamlGenerator.scala
```diff
@@ -192,46 +193,54 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
...
     td.params.size,
     defType(td),
     td.body,
     MExtern.Cpp(
-      nested(td, "cpp")("typename").toString,
-      nested(td, "cpp")("header").toString,
-      nested(td, "cpp")("byValue").asInstanceOf[Boolean]),
+      nested(td, spec.cppOutFolder.isDefined, "cpp", "typename", _.toString),
+      nested(td, spec.cppOutFolder.isDefined, "cpp", "header", _.toString),
+      nested(td, spec.cppOutFolder.isDefined, "cpp", "byValue", _.asInstanceOf[Boolean])),
```
jothepro, line 204: IntelliJ IDEA is always complaining to me that in Scala boolean parameters should be named:
```suggestion
      nested(td, isRequired: spec.cppOutFolder.isDefined, "cpp", "typename", _.toString),
      nested(td, isRequired: spec.cppOutFolder.isDefined, "cpp", "header", _.toString),
      nested(td, isRequired: spec.cppOutFolder.isDefined, "cpp", "byValue", _.asInstanceOf[Boolean])),
```

I kind of like that rule. Maybe you want to add the names?

jothepro (commented):
Cool, thank you @freitass !!
I will run the C# tests & give you feedback if they where successful.
I've left a few comments & questions, but in general this looks very good to me!

src/main/scala/djinni/Main.scala
freitass: I had thought about passing only the parts of `outSpec` that were necessary (e.g. `cppOutDir`) and call `isDefined` on them "in the last moment", but it didn't feel right. Defining the parameters to `resolve()` that way -- `<lang>OutRequired` -- definitely makes much more sense. I'd probably even leave the "out" part out: `langRequired`. Let me give it a shot and we see. Thanks for the suggestion!

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -220,5 +220,13 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
       assertFileExists(s"$outputPath/dh__map_int8_t_bool.py")
       assertFileExists(s"$outputPath/dh__set_bool.py")
     }
+
+    it("should be able to parse yaml files without all languages defined") {
+      val outputPath = "src/it/resources/result/only_yaml_out"
+      When("calling the generator with ")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
```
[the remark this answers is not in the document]
freitass: Yes, it's because of the missing `typename`.

Sounds good, they catch bugs in separate functions (methods?) indeed.

src/main/scala/djinni/YamlGenerator.scala
freitass: I do like it too. I tried using IntelliJ IDEA the other day but I didn't find an easy way to set up the project (e.g. for running the tests). I'll see if I can configure my warnings in a similar way or have IntelliJ behave :P

jothepro: That's weird, for me it worked out of the box. Maybe you need to install the scala plugin? 
For building and testing I use the integrated sbt shell in the IDE, but executing a specific test from the `>` icon on the test method should work as well.

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -220,5 +220,18 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+      val outputPath = "src/it/resources/result/only_yaml_out"
+      When("calling the generator with ")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --cppcli-out $outputPath")
+      noException should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --java-out $outputPath")
+    }
+
+    it("should gracefully fail when a required language definition is incomplete") {
+      val outputPath = "src/it/resources/result/only_yaml_out"
+      When("calling the generator with ")
+      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
```
jothepro, line 234:
```suggestion
      Given("an IDL file that depends on a YAML type definition that misses a required key in the `objc` definition")
      When("calling the generator")
      Then("the generation should fail gracefully if code for Objective-C is generated")
      a [RuntimeException] should be thrownBy djinni(s"--idl src/it/resources/date_no_cs.djinni --objc-out $outputPath")
```

jothepro (approved):
Yess, I like that a lot! Thanks! 💪 

I am proposing a more verbose test description, though. I like to define my tests in [given, when, then](https://martinfowler.com/bliki/GivenWhenThen.html) terms to make the more readable for strangers. This also comes with the bonus that the test output gives a relatively precise picture what is being tested, without even having to look at the code.

freitass:
Awesome @jothepro! Really appreciate the feedback and suggested improvements.

PR #56, opened 2021-06-07
Issue/55 oc prefix optional

### LINKED ISSUE #55: generated code doesn't compile
Currently --objc-type-prefix is an optional objc argument. If this is left off the generated objc code will not compile as there are type collisions.

Even if you move the c++ types into a namespace (--cpp-namespace test) the collision still exists, as the generated code does not specify an explicit scope when referencing the objc classes.

```
user = record {
    id: i64;
    name: string;
}
```

this will generate the following 

```
// This file was generated by Djinni from core.djinni

#import "User.h"
#include "user.hpp"

static_assert(__has_feature(objc_arc), "Djinni requires ARC to be enabled for this file");

@class User;

namespace djinni_generated {

struct User
{
    using CppType = ::test::User;
    using ObjcType = User*;

    using Boxed = User;

    static CppType toCpp(ObjcType objc);
    static ObjcType fromCpp(const CppType& cpp);
};

}  // namespace djinni_generated
```
```
// This file was generated by Djinni from core.djinni

#import "User+Private.h"
#import "djinni/objc/DJIMarshal+Private.h"
#include <cassert>

namespace djinni_generated {

auto User::toCpp(ObjcType obj) -> CppType
{
    assert(obj);
    return {::djinni::I64::toCpp(obj.id),
            ::djinni::String::toCpp(obj.name)};
}

auto User::fromCpp(const CppType& cpp) -> ObjcType
{
    return [[User alloc] initWithId:(::djinni::I64::fromCpp(cpp.id))
                                   name:(::djinni::String::fromCpp(cpp.name))];
}

}  // namespace djinni_generated
```

As you can see `using ObjcType = User*;` references User in the current scope (which is the struct), the generated code should be `using ObjcType = ::User*;` to ensure that it references the objc class (which are always in the global scope)
The same issue occurs with ` using Boxed = User;` and the 'fromCpp' implementation.

Additionally in the generated implementation file the 'fromCpp' references the Objc class directly instead of using the type alias (ObjcType)

The generator should be updated to
a) reference objc types via the global scope
b) <type>::fromCpp should use the type alias (ObjcType) instead of the type directly
c) require at least one of --objc-type-prefix --cpp-namespace to ensure the generated code is compilable.

Additional suggestion:
use a decorator suffix for the generated converter classes (UserType, UserConverter) to further disambiguate the name collisions.

a4z:
This should fix #55 

* When using  `--cpp-namespace` , ` --objc-type-prefix` becomes optional
* Without  `--cpp-namespace` , ` --objc-type-prefix` is mandatory

### DIFF
--- a/src/it/resources/expected/all_datatypes/objcpp/ITAllDatatypes+Private.h
+++ b/src/it/resources/expected/all_datatypes/objcpp/ITAllDatatypes+Private.h
@@ -13,7 +13,7 @@ namespace djinni_generated {
 struct AllDatatypes
 {
     using CppType = ::AllDatatypes;
-    using ObjcType = ITAllDatatypes*;
+    using ObjcType = ::ITAllDatatypes*;
 
     using Boxed = AllDatatypes;
 
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -188,8 +188,8 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
       val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --objc-type-prefix Cpp --objc-out $outputPath")
       Then("the generator should successfully generate just objc output")
       output should equal ("Parsing...\nResolving...\nGenerating...\n")
-      assertFileExists(s"$outputPath/AllDatatypes.h")
-      assertFileExists(s"$outputPath/AllDatatypes.mm")
+      assertFileExists(s"$outputPath/CppAllDatatypes.h")
+      assertFileExists(s"$outputPath/CppAllDatatypes.mm")
     }
 
     it("should be able to only generate Obj-C output when using --cpp-namespace") {
--- a/src/it/scala/djinni/GeneratorIntegrationTest.scala
+++ b/src/it/scala/djinni/GeneratorIntegrationTest.scala
@@ -182,15 +182,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
       assertFileExists(s"$outputPath/AllDatatypes.java")
     }
 
-    it("should be able to only generate Obj-C output") {
+    it("should be able to only generate Obj-C output with --objc-type-prefix") {
       val outputPath = "src/it/resources/result/only_objc_out"
-      When("calling the generator with just `--objc-out`")
-      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --objc-out $outputPath")
+      When("calling the generator with just `--objc-out and --objc-type-prefix Cpp`")
+      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --objc-type-prefix Cpp --objc-out $outputPath")
       Then("the generator should successfully generate just objc output")
       output should equal ("Parsing...\nResolving...\nGenerating...\n")
       assertFileExists(s"$outputPath/AllDatatypes.h")
       assertFileExists(s"$outputPath/AllDatatypes.mm")
     }
+
+    it("should be able to only generate Obj-C output when using --cpp-namespace") {
+      val outputPath = "src/it/resources/result/only_objc_out"
+      When("calling the generator with just `--objc-out and --cpp-namespace Cpp`")
+      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --cpp-namespace Cpp --objc-out $outputPath")
+      Then("the generator should successfully generate just objc output")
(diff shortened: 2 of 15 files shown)


a4z:
I do not understand that CI problem, the tests runs without problems on my machine.
This is not good

a4z:
ok, now I can reproduce, https://github.com/cross-language-cpp/djinni-generator/issues/57
will check what's going on

a4z:
ah, ok, I define a prefix that will go into the expected file names.. will fix later on the evening

a4z:
ok, now it works, but we need to fix #57 in some future, it is a really unpleasant experience having CI failing while being tricked by the local state because you need to run ` rm -r src/it/resources/result/*` before every test run

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -182,16 +182,34 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+      val outputPath = "src/it/resources/result/only_objc_out"
+      When("calling the generator with just `--objc-out and --cpp-namespace Cpp`")
+      val output = djinni(s"--idl src/it/resources/all_datatypes.djinni --cpp-namespace Cpp --objc-out $outputPath")
       Then("the generator should successfully generate just objc output")
       output should equal ("Parsing...\nResolving...\nGenerating...\n")
       assertFileExists(s"$outputPath/AllDatatypes.h")
       assertFileExists(s"$outputPath/AllDatatypes.mm")
     }
 
+    it("should not be able to generate Obj-C output when either a namespace of prefix is missing") {
```
jothepro, line 205:
```suggestion
    it("should not be able to generate Obj-C output when either a namespace or prefix is missing") {
```
?

src/it/resources/expected/all_datatypes/objcpp/ITAllDatatypes+Private.h
```diff
@@ -13,7 +13,7 @@ namespace djinni_generated {
 struct AllDatatypes
 {
     using CppType = ::AllDatatypes;
-    using ObjcType = ITAllDatatypes*;
+    using ObjcType = ::ITAllDatatypes*;
```
jothepro, line 16: I don't know enough about Objective-C++ to answer this myself without testing it:

`ITAllDatatypes` is an Objective-C datatype. You define that `ITAllDatatype` is in the root namespace (`::`), but afaik Objective-C doesn't know the concept of Namespaces. Is this syntax allowed then? Can you attest that this compiles? 🤔

jothepro (commented):
Thanks for your efforts! I am looking forward to Chris' feedback!

jothepro:
> it is a really unpleasant experience having CI failing while being tricked by the local state

Sry, I feel like not cleaning up the state goes onto my account somehow... Hope you didn't suffer too much... 🙈

src/it/resources/expected/all_datatypes/objcpp/ITAllDatatypes+Private.h
a4z: the code is Objective-C++ , note the ++ after the C ;-)  
It happens in the namespace , and getting the Objective-C data type from the global namespace is one important aspect of that PR

src/main/scala/djinni/Main.scala
```diff
@@ -361,6 +361,14 @@ object Main {
       }
     }
 
+    // Ensure either --cpp-namespace , --objc-type-prefix are given when Objective C is generated
+    if(objcOutFolder.isDefined) {
+      if (cppNamespace.isEmpty() && objcTypePrefix.isEmpty()) {
+          System.err.println("Error: At least one of [--cpp-namespace, --objc-type-prefix] needs to be set when generating Objective-C code.")
+          System.exit(1); return
```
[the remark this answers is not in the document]
a4z: In theory, yes, if some one makes a list of problem causes and exit codes, we can do that.

If, they probably should be > 2 and < 126, to not collide with bash exit codes, but other shells might have other  pre defined codes .... so it's not totally obvious trivial to me how to do that

a4z:
Thanks for the improvements!
 
> Sry, I feel like not cleaning up the state goes onto my account somehow... Hope you didn't suffer too much... 🙈

That's not important. I just had the 'luck' to got bitten by that, so I noticed it.
It could, btw, also be that it is not enough to clean the state before running tests.
If I would have had the test in different order, then I would not have noticed anything, even if one of the test implementations was incorrect, but we will go into such details in #57

paulo-coutinho:
To me appear be ok.

paulo-coutinho:
image

PR #58, opened 2021-06-12
add node support

paulo-coutinho:
Target:

- add node support
- or at least v8 compatible data


paulo-coutinho:
How to test:

```
sbt compile
sbt assembly
./run.sh
```

This is temporary and will be deleted.

paulo-coutinho:
How to check and compare with original nodejs author:

https://github.com/dropbox/djinni/compare/master...LedgerHQ:master

a4z:
Hi Paulo,  that is interesting, thanks!

It would help a lot to view over those changes when you

* Remove all files that do not belong to this PR
* Adopt the .gitignore , if required, so such files will be ignored by default
* Add some more description what this is supposed to do and how it is supposed to work
* If there are problems, what problems those are and how to reproduce
* Any other info or questions

there are, btw, other nodejs generator implementations also, at least I think @bojanin has also one 

In general, to add a new language, we need to

* have the generator adopted
* have the tests adopted
* have the support lib adopted
* have examples on how to use new language
* have the documentation adopted 

So it requires some more work to add a new binding

a4z:
I close this PR for now since in the current state there is now way this will ever be handled.
It can be re-opened any time once it is adopted the the requirements of a mergeable PR .

PR #59, opened 2021-06-12
add new option --objcpp-header-out

jothepro:
Another day, another new parameter 👏 

I propose the introduction of a new parameter `--objcpp-header-out` that complements the existing `--objcpp-out`.

## Why?

I am currently researching ways to modularize my Djinni project, [following our documentation](https://djinni.xlcpp.dev/djinni-generator/modularization/).
This is when I realized that while the Objective-C++ files generated by djinni are *private* from the end-users point of view, they are not from a developers perspective who wants to use the Djinni library in another Djinni interface.

As you can see in this excerpt of a type description YAML generated by Djinni, the private Objective-C++ header is mentioned in the interface and must be referenced by another library, if it wants to use the type:

```
name: my_record
typedef: 'record'
params: []
prefix: ""
objc:
  typename: 'DSLMyRecord'
  pointer: true
  hash: '%s.hash'
  boxed: 'DSLMyRecord'
  header: '"DjinniSecondLibrary/DSLMyRecord.h"'
objcpp:
  translator: '::Djinni::SecondLibrary::objcpp::MyRecord'
  header: '"DjinniSecondLibrary/DSLMyRecord+Private.h"'
```

## User Story

As a developer I want to be able to generate the Objective-C++ header files in a separate directory from the Objective-C++ sources. This makes it easier for me to distribute a binary with all public header files that are required to use the library in another Djinni library [with `@extern`](https://djinni.xlcpp.dev/djinni-generator/modularization/#usage).

Im am open for discussion if this is really necessary, my current workaround is to filter the Objective-C++ output for all header files and install them, which is not too painful after all. But the introduction of the parameter would make my life easier.


src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -67,34 +70,36 @@ class IntegrationTest extends FunSpec {
...
+      s"--objcpp-header-out $baseOutputPath/$idl/$OBJCPP_HEADERS " +
+      s"--py-out $baseOutputPath/$idl/python " +
+      s"--pycffi-out $baseOutputPath/$idl/cffi " +
+      s"--c-wrapper-out $baseOutputPath/$idl/$CWRAPPER " +
+      s"--c-wrapper-header-out $baseOutputPath/$idl/$CWRAPPER_HEADERS " +
+      s"--c-wrapper-include-prefix ../$CWRAPPER_HEADERS/ " +
+      s"--c-wrapper-include-cpp-prefix ../$CPP_HEADERS/ " +
+      s"--cppcli-out $baseOutputPath/$idl/$CPPCLI " +
+      s"--cppcli-include-cpp-prefix ../$CPP_HEADERS/ " +
+      s"--list-out-files $baseOutputPath/$idl/generated-files.txt " +
```
jothepro, line 94: Don't be overwhelmed by these changes. I realized that I could/should make use of the existing Language variables (`CPP`, `CPP_HEADERS`, ...), which specify where to find the generated output. 
The only real change in here is the addition of
```
s"--objcpp-header-out $baseOutputPath/$idl/$OBJCPP_HEADERS " +
```

src/main/scala/djinni/ObjcppGenerator.scala
```diff
@@ -378,7 +378,8 @@ class ObjcppGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
  }
 
   def writeObjcFile(fileName: String, origin: String, refs: Iterable[String], f: IndentWriter => Unit) {
-    createFile(spec.objcppOutFolder.get, fileName, (w: IndentWriter) => {
+    val folder = if (fileName.endsWith(spec.objcHeaderExt)) spec.objcppHeaderOutFolder else spec.objcppOutFolder
```
jothepro, line 381: Choosing the output folder depending on the file extension seems a bit fragile to me. I feel like it would be better to explicitly tell `writeObjcFile` if it is currently writing a header or source file.

But this seems to be how it's done in other places too, so I decided to go the easy way that keeps this PR small. Maybe maybe this should eventually be refactored somewhere in the future, what you think?... 🙈

a4z:
That is an interesting use case, I never worked with djinni yaml part.

If the headers could be required, it should be possible to put them under an include prefix like the other public headers.
You are totally right. Thanks for your suggestion!
After all, the _+Private_ part is mostly a ObjectiveC extension thing, not that this header is really private. We could also choose n different extension name.

I will look over it more closely this evening, but after a fist view, lgtm!

src/main/scala/djinni/ObjcppGenerator.scala
a4z: it might be possible to catch that more early, 
https://github.com/cross-language-cpp/djinni-generator/blob/b01a3069e0927f5e0fd681ec13dd20c0150344ea/src/main/scala/djinni/ObjcppGenerator.scala#L338 for the header

for the implementation, https://github.com/cross-language-cpp/djinni-generator/blob/b01a3069e0927f5e0fd681ec13dd20c0150344ea/src/main/scala/djinni/ObjcppGenerator.scala#L357

at least, from reading it looks like that

if the whole file name would be generated there, and `writeObjcFile` would only write to the filename it gets, without applying it's own magic, what should be the case anyway, than that could work

would you like to give that a trial?

src/it/scala/djinni/IntegrationTest.scala
a4z: yes, it's just we do not have any tests for the default parameter, right? 
not that this is important, I can see in the code that that should work (as long as no one changes the order of things)
but the list of command line option cases we do not test also grew :-)

jothepro: You are right about that. I added a special test case that is making sure the `--objcpp-out` parameter is working standalone. While I am aware that this is only implicitly tests the fallback, I think it will do the job. I tried to make it clear in the test description that the fallback is also something that we want to make sure...
https://github.com/cross-language-cpp/djinni-generator/blob/c4768384cf29aad06109f96fae502141b5600d90/src/it/scala/djinni/GeneratorIntegrationTest.scala#L224

src/main/scala/djinni/ObjcppGenerator.scala
jothepro: I decided to introduce a parameter `isHeader`, that tells `writeObjCFile` if what it writes is supposed to go into the header or the source dir. It was pretty straightforward, I may introduce the same concept in other places as well in a separate PR.

a4z: Not sure about `isHeader`, can you please explain a little bit more

To me it seems, when I can pass `isHeader` and a file name, I could also directly construct the file path.
That would eliminate a parameter, a if branch, and let writeFile make what it should , writing a file, not guessing a directory.
Or is that not possible?

jothepro: The problem is that the filename is constructed differently in every call of `writeObjCFile`:

L56: `objcppMarshal.privateHeaderName(ident.name)`
L96: `objcppMarshal.privateHeaderName(ident.name)`
L145: `privateBodyName(ident.name)`
L338: `objcppMarshal.privateHeaderName(objcName)`
L357: `privateBodyName(objcName)`

This is how `objcName` is defined: `val objcName = ident.name + (if (r.ext.objc) "_base" else "")`

I didn't find a pattern in this that could easily be moved into the write function, that's why the name is still passed in by the caller.

a4z: I think we can assume that the places where `privateBodyName` is used are the implementation files, `mm`, the other the header.
But we put this as an extra topic for the future, we could also review the same in the objectivec part, where it is exactly the same.

Thanks for the help understanding the code base better!

I think it is possible to simplify that in some future, but please proceed as you do, or as you have it, since we should not mix too much into that ticket that solves a very specific problem, and that it does well. (and pls feel free to resolve that conversation after reading this)

a4z (approved):
Great addition! Thanks!!

src/main/scala/djinni/ObjcppGenerator.scala
jothepro: ohh, now I get it... 🤦  `objcppMarshal.privateHeaderName()` and `privateBodyName()` could indeed be moved into `writeObjCFile`. I can then just pass in `ident.name` or `objcName` according to the context. 🙈 
Thx, I'll do that. 

> we could also review the same in the objectivec part

Yep, it's the same there so it should be refactored, but not in this PR 👍

a4z (approved):

PR #60, opened 2021-06-13
ObjcGenerator: move `fileName` generation calls into `writeObjcFile`

jothepro:
As discussed I am making the same adjustments to the `ObjcGenerator` that where done to the `ObjcppGenerator`.


src/main/scala/djinni/ObjcGenerator.scala
```diff
@@ -191,7 +191,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
         w.wl
         writeDoc(w, f.doc)
         val nullability = marshal.nullability(f.ty.resolved).fold("")(", " + _)
-        w.wl(s"@property (nonatomic, readonly${nullability}) ${marshal.fqFieldType(f.ty)} ${idObjc.field(f.ident)};")
+        w.wl(s"@property (nonatomic, readonly$nullability) ${marshal.fqFieldType(f.ty)} ${idObjc.field(f.ident)};")
```
a4z, line 194: is this change an accident or on purpose, and if wanted, what does it improve?

jothepro: That's not an accident, but unrelated to this PR. The curly brackets are not required if passing just a variable to a string template. The IntelliJ Linter complained about this when committing my changes. The change does not improve anything but the syntax.

a4z: Thanks for the explanation! 

IMHO, readability wise, the linter should complain exactly about the opposite, if the {} are missing. 
Can't it be configured ? 

Anyhow, this should not block this PR, but in general. I always prefer better readable solutions. If a linter is not possible to support that for no technical reason other than just because, we should not listen to it.

jothepro:
```suggestion
        w.wl(s"@property (nonatomic, readonly${nullability}) ${marshal.fqFieldType(f.ty)} ${idObjc.field(f.ident)};")
```
I actually do ignore a lot of linter warnings by the IDE if I don't share the same opinion or if I don't care. I did not want to reduce readability. I'm fine with rolling this back if you don't like it :)

a4z (approved):
Thanks !

PR #62, opened 2021-06-16
Support non-null external interfaces

### LINKED ISSUE #61: Non-null pointers do not work when using modularization
Using non-null pointer types with the --cpp-nn-header argument together with modularization does not work properly.

Here’s a toy example:
a.djinni
```
a = interface +c {
    const test();
}
b = interface +c {
    const test(a: a): a;
}
```

c.djinni
```
@extern "a.yaml"
c = interface +c {
    const test(a: a): a;
}
```

run-djinni.sh
```
#! /usr/bin/env bash
set -e
module="$1"
base_dir=$(cd "`dirname "0"`" && pwd)
cpp_header_out="$base_dir/cpp/include/abc/$module"
cpp_src_out="$base_dir/cpp/src/$module"
namespace="abc"
djinni_file="$module.djinni"
djinni \
  --cpp-out $cpp_src_out \
  --cpp-header-out $cpp_header_out \
  --cpp-namespace $namespace \
  --cpp-nn-header "\"nn.hpp\"" \
  --cpp-nn-type "dropbox::oxygen::nn_shared_ptr" \
  --cpp-nn-check-expression "NN_CHECK_ASSERT" \
  --yaml-out . \
  --yaml-out-file ${module}.yaml \
  --idl $djinni_file
```

Executing `run-djinni.sh` for both modules `a` and `c` generates the three headers `a.hpp`, `b.hpp` and `c.hpp`:
```
./run-djinni.sh a
./run-djinni.sh c
```

cpp/include/abc/a/a.hpp
```
// This file was generated by Djinni from a.djinni
#pragma once
namespace abc {
class A {
public:
    virtual ~A() {}
    virtual void test() const = 0;
};
}  // namespace abc

```

cpp/include/abc/a/b.hpp
```
// This file was generated by Djinni from a.djinni
#pragma once
#include "nn.hpp"
#include <memory>
namespace abc {
class A;
class B {
public:
    virtual ~B() {}
    virtual dropbox::oxygen::nn_shared_ptr<A> test(const dropbox::oxygen::nn_shared_ptr<A> & a) const = 0;
};
}  // namespace abc
```

cpp/include/abc/c/c.hpp
```
// This file was generated by Djinni from c.djinni
#pragma once
#include "a.hpp"
#include <memory>
namespace abc {
class C {
public:
    virtual ~C() {}
    virtual std::shared_ptr<::abc::A> test(const std::shared_ptr<::abc::A> & a) const = 0;
};
}  // namespace abc
```
I think ```C::test``` should take and return a ```shared_ptr<A>``` rather than a ```nn_shared_ptr<A>```  just like ```B::test``` does.

roman-captural:
- Fixes #61 by adding support for non-null external interface references in the C++ generator.
- Add an integration test for the above mentioned fix.
- Parametrize construction of djinni CLI params for integration tests.


a4z (approved):
Great work, thanks a lot! 

Fixes #61 

And the encapsulation for the test djinni runner is also a valuable contribution for the future of testing, awesome.

roman-captural:
Thanks for the quick review @a4z. I found a small typo in the docs (Pyhton -> Python), but will wait for the other outstanding review in case there will be more things to fix.

src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -64,35 +64,76 @@ class IntegrationTest extends FunSpec {
...
+    *
+    * @param idl filename of the djinni-file (without file extension). The file must be located in the
+    *            `resources`-folder of the integration-tests (`src/it/resources`)
+    * @param baseOutputPath The root folder for the outputs to be generated.
+    * @param cpp Whether to generate C++ output. Default: true.
+    * @param java Whether to generate Java output. Default: true.
+    * @param objc Whether to generate Objective C output. Default: true.
+    * @param python Whether to generate Pyhton output. Default: true.
+    * @param cWrapper Whether to generate Pyhton output. Default: true.
+    * @param cppCLI Whether to generate Pyhton output. Default: true.
```
jothepro, line 77:
```suggestion
    * @param python Whether to generate Python output. Default: true.
    * @param cWrapper Whether to generate Python output. Default: true.
    * @param cppCLI Whether to generate Python output. Default: true.
```

I spotted the typos! ;)

src/it/resources/expected/cpp_dependent_interface/cpp-headers/dependent_interface.hpp
```diff
@@ -0,0 +1,21 @@
...
+public:
+    virtual ~DependentInterface() {}
+
+    virtual void method_taking_interface_dependency(const dropbox::oxygen::nn_shared_ptr<::InterfaceDependency> & dep) = 0;
+
+    virtual void method_taking_optional_interface_dependency(const std::shared_ptr<::InterfaceDependency> & dep) = 0;
+
+    virtual dropbox::oxygen::nn_shared_ptr<::InterfaceDependency> method_returning_interface_dependency() = 0;
+
+    virtual std::shared_ptr<::InterfaceDependency> method_returning_optional_interface_dependency() = 0;
```
jothepro, line 20: Can you help me understand this part? My naive understanding is that that `method_taking_optional_interface_dependency` and `method_returning_optional_interface_dependency` should take/return a `std::optional` and not just a plain `shared_ptr`?

jothepro (commented):
Thanks for your contribution! And thanks for providing a test!
I am not very familiar with the concepts of non-null & optional in djinni, that's why I have one question about the expected test outcome.

src/it/resources/expected/cpp_dependent_interface/cpp-headers/dependent_interface.hpp
roman-captural: `std::shared_ptr` is an optional type already in the sense that it can point to null. That's actually also the reason that `nn_shared_ptr` is useful in djinni: To explicitly state at compile time that a pointer is not null or in other words not optional. It's handy if we don't want to check our interface types for null wherever we use them.

An `optional<my_interface>` param is always translated to `std::shared_ptr<MyInterface>`. A non-optional `my_interface` param is translated to `std::shared_ptr<MyInterface>` as well, except if we use the `--cpp-nn-header` option. In this case, with the fix in this PR, it's now correctly translated to `nn_shared_ptr<MyInterface>`.

jothepro (approved):

PR #64, opened 2021-06-17
Transform supported types List to table to improve readability

### LINKED ISSUE #63: Documentation about <optional> needs update
The documentation still says:

> Optionals (optional<typeA>). This is std::experimental::optional<T> in C++11

https://djinni.xlcpp.dev/djinni-generator/idl/#data-types

[C++11 is not supported any more](https://github.com/cross-language-cpp/djinni-support-lib/blob/main/CMakeLists.txt#L62) & the [default is `std::optional` now](https://github.com/cross-language-cpp/djinni-generator/blob/main/src/main/scala/djinni/Main.scala#L35). The documentation needs to be updated.

jothepro:
fixes #63 

This is a documentation cleanup that I had in mind for some time now. I transformed the documentation of supported types from a list to a table, as it has grown over time & was a bit hard to read.

The result looks like this:

Bildschirmfoto 2021-06-17 um 22 29 01

Because the table got quite wide, it is scrollable to the right to show the types for Python & C#:

Bildschirmfoto 2021-06-17 um 22 29 30


a4z:
Looks beautiful! 

I wonder, how complicated would it be to have the table selectable for languages, 
so you can say, show only Java and Python, for example.

Do you know if that could be done with CSS only, or would it require java script?

docs/idl.md
```diff
@@ -74,28 +74,30 @@ will only be processed once.
...
...p;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Java<br>(Java boxed)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Objective-C<br>(Obj-C boxed)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | C#&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Python&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; |
```
a4z, line 77: are all the `&nbsp` required, can we not use some css flex or so, to define minimum width ? 
I guess that should be possible

jothepro:
I did some research if there is a plugin for mkdocs for wide tables. With some kind of filter functionality or sth like that. I couldn't find sth in the official plugin list: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins

I do think that hiding/showing table rows requires some amount of JS 🤔

jothepro:
Hmm... I actually think one could abuse mkdocs content tabs: https://squidfunk.github.io/mkdocs-material/reference/content-tabs/

by having a table for each target language and a tab selector on top 🤔

a4z (approved):
A the end everything seems to be possible by providing own templates, question is just, would it be worth the effort ....

Let's see, keep that idea in mind and if time and energy allows it, explore that space.
Until then we take every improvement we can get so start with this and see what is possible over time

jothepro:
Woow, this is so much better:

Bildschirmfoto 2021-06-17 um 23 08 38

The only problem is that the width of the tables is not always the same:

Bildschirmfoto 2021-06-17 um 23 09 09

What you you think?

a4z:
Great improvement! 
I :heart: it coming in the morning to the Computer, seeing over night magic has happened :-) 

Don't think it's an issue that tables do change width and number of columns, that just how it is

jothepro:
And I like to work at night because it's soo hot here during the day :D
I think that I still have your approval to merge this?

a4z (approved):
Yes, but I can approve again :-)

PR #68, opened 2021-07-10
Suppress depreciation warning for Java finalze call

### LINKED ISSUE #16: Add @SuppressWarnings("deprecation") annotation to generated finalize methods
The use of `finalize` is a big can of worms (see StackOverflow [Why is the finalize() method deprecated in Java 9?](https://stackoverflow.com/questions/56139760/why-is-the-finalize-method-deprecated-in-java-9)), but it would be nice to at least have a djinni option to have the generated Java classes annotate the `finalize` generated methods with `@SuppressWarnings("deprecation")` so I don't get Java compile warnings.

a4z:
Fixes #16

Adds a @SuppressWarnings("deprecation") to the generated finalize call.

Since there seem not to be a way to live without the finalize methode,
we should not inject warnings into use code for that.


jothepro (approved):
👏

PR #69, opened 2021-07-10
clean test output dir at the begin of test run

### LINKED ISSUE #57: Test should clean up the result folder before running
Tests depend on what a test run generated into the it/resources/result folder.

But it is possible that there are files from previous runs, so tests on local machine require human intervention before running tests. 
It would be better to invalidate or remove all already existing files before a test is running

a4z:
Adds functionality to clean a directory with generated files, and does so at the begin of a test run.

This should fix #57 , 
It might be that there is still a dependency between tests
and a test might generated files a following tests checks,
but until we detect such a case, this should be 'good enough'

### DIFF
--- a/src/it/resources/.gitignore
+++ b/src/it/resources/.gitignore
@@ -1,0 +1,2 @@
+# Everything generated from tests
+result/*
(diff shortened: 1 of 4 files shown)


jothepro (approved):
Thx!

PR #70, opened 2021-07-11
fix link to djinni generator documentation

jothepro:
I found out by accident that the link to the djinni-generator documentation is broken. Looks like github changed sth in their gh-pages routing logic?


a4z (approved):
Good catch, thanks!

PR #71, opened 2021-07-17
Add bridging header to generated files list

### LINKED ISSUE #54: `--list-out-files` does not include `--objc-swift-bridging-header`
When the generator is run like this:

```
djinni --objc-out ... --cpp-out ... --objc-swift-bridging-header bridging-header --list-out-files list-out-files.txt
```

The file `list-out-files.txt` does not include the generated bridging header `bridging-header.h`.

As a user I'd like the bridging header to be part of the generated files list because I rely on the list as input for the `PUBLIC_HEADER` CMake property. For that I take the content of `list-out-files.txt`, parse it to a CMake list and `FILTER` it for objc-headers. In the current state I have to manually add `bridging-header.h` to this list afterwards.

a4z:
Fixes #54

If the option `--list-out-files` is used, the generated bridging header
will be included (if one is generated)


jothepro (approved):
Awesome, thx! 👏

PR #72, opened 2021-07-18
Add docs for #26

### LINKED ISSUE #26: bring generator to homebrew
As a mac user I think it would be super cool if I could just install the djinni-generator from homebrew.

a4z:
#26 was closed by a new repo I created, but it should had been closed by that PR.
(Sorry for that surprise, I was not aware I can close issues via commit message cross repo wise) 

This is the doc for the addition, installing the generator via homebrew. Closes #26 

**Pls note**, this will add a stage to the release of a new generator version, we need to update the cask, 
(until we figure out how autoupdate works)

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -26,7 +26,7 @@ We have a brew tap!
 Add the tap:
 
 ```bash
-brew cross-language-cpp/brew https://github.com/cross-language-cpp/brew.git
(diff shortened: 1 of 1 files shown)


jothepro (approved):
Cool! I've not looked into the tap sources yet but installation works on my machine 👍

Issue #44, opened 2021-04-17
Unit-Tests in djinni-support-lib can not be executed without a generator release.


jothepro:
In #43 a big new feature is introduced: Python support for djinni.

Because djinni has been split into components, we do now face a problem while introducing the big new feature that includes big changes in both generator & support-lib:

The changes in the generator need to be released, otherwise it cannot be pulled as dependency in the support-lib to execute the unit tests. Once the generator is released, the unit-tests in the support-lib repository can test if the generated code compiles & works.
If theses test fails because of an error in the generator implementation, a hotfix needs to be made in the generator and the fix has to be released again.

From my point of view this should not be a problem in general, as all things should be tested locally before being pushed to the generator/support-lib repository anyways.The tests in github action should just be there to verify that everything is fine.

But in reality things go wrong, and situations like this may get nasty, if multiple patches need to be made in the generator for some unexpected reason. As a maintainer I think we should discuss options on how to reduce this hard dependency when introducing big features that require changes in both components.

a4z:
To start the discussion, here something, maybe we can iterate on it

1) a merge into master generate a djinni binary, djinni-latest , and this is used
since this is tied to a commit, and the commit sha could be reflected in the version, there is some idea what this includes, if this info is ever needed.
And having a latest,  rolling release, available is common practice. Use at your own risk.

2) PRs for support lib has some way to know which generator to use. Which way this is, I do not know at the moment ...

jothepro:
Theoretically you can always download a canary build from the latest github actions run: 

<img width="500" src="https://user-images.githubusercontent.com/21294002/115145986-e1261b00-a054-11eb-826f-fa3ea72f1623.png">

https://github.com/cross-language-cpp/djinni-generator/actions/runs/760704611

Each "CI" workflow creates an artifact with the djinni binary.

But it may be hard to get the artifact by using the api, afaik there is no endpoint for the "latest" action/artifact.

a4z:
I think we should move that into the discussion part, since it does not seem there will be any action in reasonable time.
Also, the problem is not really a problem that stops something, it is more an inconvenience, that delays the pipeline if we run into the situation that there is a dependency. 

I think what we actually would need might be something like that
https://docs.gitlab.com/ee/ci/pipelines/multi_project_pipelines.html
and/or that
https://docs.gitlab.com/ee/ci/yaml/#cross-project-artifact-downloads-with-needs

not sure if gh action provides something similar

PR #78, opened 2021-08-26
Fix some small issues with C++/CLI code generation

freitass:
We're updating our Djinni to use a version from this repository instead of our own fork from the original repository from Dropbox. Here are the fixes for a few issues we found along the way:

- Missing namespace specification in `fqTypename` (detected by an compilation error in a `toCpp` generated code)
- ~Unreasonable defaults for the CLI params `--<lang>-base-lib-include-prefix` (values are redundant to what is already exported by the support-lib, as you can see in this Obj-C example https://github.com/cross-language-cpp/djinni-support-lib/blob/fbdec2aa6dc935e3381ebb2ed53d3bad58a84bb5/CMakeLists.txt#L79)~ (Moved to https://github.com/cross-language-cpp/djinni-generator/pull/82)


docs/cli-usage.md
```diff
@@ -87,7 +87,7 @@ djinni \
 | `--jni-include-prefix <prefix>` | The prefix for `#includes` of JNI header files from JNI C++ files. |
 | `--jni-include-cpp-prefix <prefix>` | The prefix for `#includes` of the main header files from JNI C++ files. |
 | `--jni-namespace ...` | The namespace name to use for generated JNI C++ classes. |
-| `--jni-base-lib-include-prefix ...` | The JNI base support library's include path (default: `djinni/jni/`). |
+| `--jni-base-lib-include-prefix ...` | The JNI base support library's include path. |
```
a4z, line 90: I think we just had that  due to backward compatibility, and since we changed the include to the whole path (e.g. djinni/jni/Marshal.hpp) it's defacto not required.
I just wonder, shall we not remove the option, or is there still a use case for it?

docs/cli-usage.md
```diff
@@ -115,7 +115,7 @@ djinni \
 | `--cpp-extended-record-include-prefix <prefix>` | The prefix path for `#include` of the extended record C++ header (`.hpp`) files |
 |`--objc-extended-record-include-prefix <prefix>` | The prefix path for `#import` of the extended record Objective-C header (`.h`) files  |
 |`--objcpp-namespace <prefix>` | The namespace name to use for generated Objective-C++ classes.  |
-|`--objc-base-lib-include-prefix ...` | The Objective-C base support library's include path (default: `djinni/objc/`).  |
+|`--objc-base-lib-include-prefix ...` | The Objective-C base support library's include path.  |
```
a4z, line 118: same comment as for jni applies here

docs/cli-usage.md
```diff
@@ -129,7 +129,7 @@ djinni \
 | `--c-wrapper-header-out <out-folder>` | The output folder for C wrapper header files (default: the same as `--c-wrapper-out`). |
 | `--c-wrapper-include-prefix <prefix>` | The prefix for #includes of C wrapper header files from C wrapper C++ files. |
 | `--c-wrapper-include-cpp-prefix <prefix>` | The prefix for #includes of C++ header files from C wrapper C++ files. |
-| `--c-wrapper-base-lib-include-prefix <prefix>` | The C wrapper base support library's include path (default: `djinni/cwrapper/`). |
+| `--c-wrapper-base-lib-include-prefix <prefix>` | The C wrapper base support library's include path. |
```
a4z, line 132: same comment as for jni applies here

src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -97,12 +97,14 @@ class IntegrationTest extends FunSpec {
       cmd += s" --java-out $baseOutputPath/$idl/$JAVA"
       cmd += s" --jni-out $baseOutputPath/$idl/$JNI"
       cmd += s" --jni-header-out $baseOutputPath/$idl/$JNI_HEADERS"
+      cmd += " --jni-base-lib-include-prefix djinni/jni/"
```
a4z, line 100: this should not be needed, we do not want to force users to specify include directories that should  just be found.

src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -97,12 +97,14 @@ class IntegrationTest extends FunSpec {
...
     }
     if (objc) {
       cmd += s" --objc-out $baseOutputPath/$idl/$OBJC"
       cmd += s" --objc-header-out $baseOutputPath/$idl/$OBJC_HEADERS"
       cmd += " --objc-swift-bridging-header bridging-header"
       cmd += " --objc-type-prefix IT"
+      cmd += " --objc-base-lib-include-prefix djinni/objc/"
```
a4z, line 107: same as above

src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -115,10 +117,12 @@ class IntegrationTest extends FunSpec {
       cmd += s" --c-wrapper-header-out $baseOutputPath/$idl/$CWRAPPER_HEADERS"
       cmd += s" --c-wrapper-include-prefix ../$CWRAPPER_HEADERS/"
       cmd += s" --c-wrapper-include-cpp-prefix ../$CPP_HEADERS/"
+      cmd += " --c-wrapper-base-lib-include-prefix djinni/cwrapper/"
```
a4z, line 120: same as above

src/it/scala/djinni/IntegrationTest.scala
```diff
@@ -115,10 +117,12 @@ class IntegrationTest extends FunSpec {
...
     }
     if (cppCLI) {
       cmd += s" --cppcli-out $baseOutputPath/$idl/$CPPCLI"
       cmd += s" --cppcli-include-cpp-prefix ../$CPP_HEADERS/"
+      cmd += " --cppcli-base-lib-include-prefix djinni/cppcli/"
```
a4z, line 125: same as above

src/main/scala/djinni/Main.scala
```diff
@@ -57,7 +57,7 @@ object Main {
     var jniIncludeCppPrefix: String = ""
     var jniFileIdentStyleOptional: Option[IdentConverter] = None
     var jniBaseLibClassIdentStyleOptional: Option[IdentConverter] = None
-    var jniBaseLibIncludePrefix: String = "djinni/jni/"
+    var jniBaseLibIncludePrefix: String = ""
```
a4z, line 60: same as in the context of the other related include paths....

a4z (changes requested):
That feels kind of wrong.

Please help me to find out if I understand wrong.

We added the generation of  full include paths`#include "djinni/jni/Marshal.hpp"` to not require specify them. 
If the support lib is in an include path, the generated code shall just compile because it finds the header, because the include is `djinni/binding/whatever.h`

This change seems to destroy that, and requires going back to a state where we need to specify each include for each binding on the command line, what is imho unwanted.

We should have sane defaults. and that's it., 

That the cmake of the support lib has an explicit build interface is not relevant if the support lib is not a sub project, but a stand alone package.

I think the right way to move forward is to remove those options at all, and generate what we know the support lib looks like, because I do not see any value in those options. And for sure they have none if users need to put them on the command line.

docs/cli-usage.md
freitass: That's a good question. In our project we have the support lib as a submodule, so the include path is flexible. I don't think it's much different if you're installing the support lib instead. To be honest, I tend to agree it's a good opportunity to clean up some of these options. Should we create a separate PR and get some feedback from the community on Slack before removing those?

freitass:
@a4z I fully agree with your assessment. I'll try to remove those options to evaluate the impact on the support lib.

a4z:
Awesome changes, Leandro, thanks a lot! 

Just one thing, we have here 2 fixes in one PR, and the PR title doe not match anymore.

I wonder if it would be possible to split it into 2 PRs, one with removal of the prefix command line args, the other with the original title (small fix of cpp/cli issue) 

especially since the command line args removal has a follow up in https://github.com/cross-language-cpp/djinni-support-lib/pull/53 and relates to #44 since I think that means the support lib needs a new generator release, and having that split would allow us to thing about the problem without ignoring (or delaying) the C++/CLI fix

freitass:
@a4z Good idea! I've moved the removal of the flags to https://github.com/cross-language-cpp/djinni-generator/pull/82.

a4z (approved):
You are awesome Leandro, thanks a lot!

jothepro:
@freitass can you help me understand what effect this change has? Maybe we could add a test for it?

freitass:
@jothepro this issue happened when you have a property in a record that has the same name as its enum type, e.g.:
```djinni
my_enum = enum {
  a;
}

my_record = record {
  my_enum: my_enum;
}

```
In C++/CLI generated code, `ToCpp` would be as follows (approx.):
```cpp
MyRecord::CppType MyRecord::ToCpp(MyRecord::CsType cs)
{
  ASSERT(cs != nullptr);
  return {::djinni::Enum<::cpp::namespace::MyEnum, MyEnum>::ToCpp(cs->MyEnum)};
}
```
Notice the second template parameter in `::djinni::Enum<::cpp::namespace::MyEnum, MyEnum>::ToCpp(cs->MyEnum)` -- "MyEnum" refers to the property rather than the type. The same thing happens in `FromCpp`.

I've submitted a PR to cover this case on the generator tests: https://github.com/cross-language-cpp/djinni-generator/pull/85

PR #79, opened 2021-08-26
Generate a default constructor for records in C++ unless 'omit' flag is specified

freitass:
We're in the process of updating our Djinni to use a version from this repository instead of our own fork from the original repository from Dropbox. While some of the changes we had introduced in our fork were already integrated in this repository (namely the support for C++/CLI), some others are still pending and we'll either submit new PRs as we need them or end up dropping them (if they aren't required at all).

In this PR I'm proposing the introduction of a flag `--cpp-generate-default-record-constructor` that allows Djinni to generate an additional `= default` constructor in C++ records. This proposal is inspired in the discussion that took place at https://github.com/dropbox/djinni/issues/113 and it could be later combined with a more finer grained control through extensions. Our main motivation to have this feature is when using Djinni records with `std::promise`; there's currently a bug in MSVC that requires the types used as template parameters to be default constructible.

Please note the constructor is being defined as `= default` and I state it explicitly in the docs. And I'm open to suggestions for a shorter/better cli flag.


a4z:
There is a lot of discussion in the original Dropbox repo on that topic, and I think some of them have a good point.
Especially the lack of defined init values for the case when the default constructor is called.

That said, I have self being in the situation that I wanted a default constructor, since it had also a situation where constructing records with values was just boiler plate, since those records have been overwritten soon anyway, and I wished that would have been possible. (I had actually even some kind of patch for that...)

I just wonder if we should explore that topic a little bit more in depth, like, providing user the chance to  specify default values, like statics already do. (or at least having a `{}` behind the record fields so they are default constructed them self, but I would need to read docs to check if that is the same in some cases)

a4z (changes requested):
I found the reference I was thinking about
 
https://arne-mertz.de/2015/08/new-c-features-default-initializers-for-member-variables/

```
If class members are neither mentioned in a constructor’s member initializer list nor have a brace-or-equal-initializer, then they get default-initialized. That means, that for class types the default constructor is called, but for any other types like enums or built in types like int, double, pointers, no initialization happens at all.
```

in short, I think we should add `{}` initizlizer to the field (record) member , then it should be fine and we avoid garbage values

freitass:
@a4z trying not to break too much the existing patterns in the code, I'm listing all members in the constructor initializer list. According to this paragraph from the link you posted it has the same effect:

> If class members are neither mentioned in a constructor’s member initializer list nor have a brace-or-equal-initializer, then they get default-initialized. That means, that for class types the default constructor is called, but for any other types like enums or built in types like int, double, pointers, no initialization happens at all.

Here's what the resulting code looks like:

```cpp
    MyRecordDefaultCtor()
    : id()
    , info()
    , store()
    , hash()
    {}
```

This is in fact what we've done at some point in our own branch, so it was also easier for me to pull it in... 😅

a4z (commented):
Thanks Leandro! 

I just wonder, should we not make that just the new normal?
Without adding a new flag `--cpp-generate-default-record-constructor` ? 
We have already enough flags, and they are expensive to maintain.

So maybe we flip the logic of the new flag, say `--cpp-omit-default-record-constructor` so we can generate code as it was, for a while, and then sunset that flag.

Opinions?

freitass:
I love that idea! I had also forgotten to update the docs regarding the `= default` ctor.

a4z (approved):
Awesome, thanks a lot Leandro!

PR #80, opened 2021-09-03
Only initialize the objcSwiftBridgingHeaderWriter if generating files

eakoli:
The Swift bridging header would be created as a 0 byte file of —skip-generation true was passed.


a4z:
Hi Chris, thanks a lot for your fix !

Do you think it is possible to add a test that handles the case (skipping shall not generate a file)

a4z (changes requested):
I would like to have that change, but we need a unit test for ot.
If you can add one, @eakoli , that would be great!
If not, you need to wait until one of us has time to update the PR with according tests

a4z (approved):
Awesome, thanks a lot @eakoli !

PR #81, opened 2021-09-06
Clear references to Python from C wrapper

freitass:
Make function/method and variable names generated for the C wrapper more generic (independent from Python). A few references to Python remain in Scala code and generated C/C++ code, such as these annotations in generated C header files. The comments below seem to tell CFFI to ignore those headers but a Google search doesn't return any results for that string.
```cpp
#pragma once // python_cdef_ignore
#include <stdbool.h>  // python_cdef_ignore

#include <stdbool.h> // python_cdef_ignore
#include <stdint.h> // python_cdef_ignore
```

The generated proxies for Python object are also good candidates for renaming, even though I still left named as `FooPythonProxy`. The reason is because I'm looking for a good name that refer to any language that could replace Python in using the C wrapper.


a4z:
no idea about the `python_cdef_ignore`, I wish I could find the time to work more on the pthon bindings.

The PR looks good, only thing I wonder is, when touching that code, should we not also cleanup the double and tripple `_` in the names, and have everything with  just 1 underscore ? 

examples with `_` are, for example
`list_bool_add_callback__create` 
`s_callback_all_datatypes___delete` 
but there are many more

a4z (approved):

a4z:
After thinking a little more, I think the double and tripple `_` are ok, this is generated code, it is supposed to look different, and it makes searching for symbols more easy.

freitass:
Indeed, I see it as a way to make a clear separation between the components of the name. I was looking for some C coding guideline that would suggest that kind of thing but I wasn't able to find any. I'm glad you also saw value on them. :)

freitass:
Sorry @a4z, I'm requesting your review once more because I added another commit: https://github.com/cross-language-cpp/djinni-generator/pull/81/commits/58314a80afbff9bcec141afc89acfec5722c54d1

I replaced the remaining instances of "python" with "front-end".

a4z:
my existing approval has not been invalidated, and I still think that is fine.
Just wonder if any of those changes requires updates in the support lib, but I guess we will find out :-)

freitass:
I've been testing with no modifications to the support lib and it still works :)

PR #82, opened 2021-09-08
Remove all CLI params --<lang>-base-lib-include-prefix

freitass:
Defaults values for the CLI params --<lang>-base-lib-include-prefix are currently redundant to what is already exported by the support-lib, as you can see in this Obj-C example https://github.com/cross-language-cpp/djinni-support-lib/blob/fbdec2aa6dc935e3381ebb2ed53d3bad58a84bb5/CMakeLists.txt#L79. This PR proposes the removal of those parameters in an effort to clean up the command line interface.

(PR split from https://github.com/cross-language-cpp/djinni-generator/pull/78)


jothepro (approved):
I was tempted to dislike the idea at a first glance because it is introducing a breaking change without a real value other than cleaning up the CLI interface, but I've changed my mind. The next release of the generator probably will be a major version jump anyways, considering the new default constructors. 💪 Let's do this!

jothepro:
Just for reference, this PR is related to https://github.com/cross-language-cpp/djinni-support-lib/pull/53

a4z (approved):

PR #83, opened 2021-09-13
add missing comments to generated output

### LINKED ISSUE #67: include comments from IDL in generated C++/CLI
When a djinni interface is defined with comments:

```
# comment
Foo = interface +c +j +o +s {
  bar();
}
```

The comments are not copied to the generated C++/CLI class.

As a user I'd like the comments to be copied to the C++/CLI class like it happens with all other languages as well. That way I can generate a documentation of the generated C++/CLI code that documents the interfaces with the comments from the djinni IDL file.

jothepro:
fixes #67 

Comments are now also copied to C++/CLI interfaces.
While adding comments to the integration-test IDL files to make sure everything works, I realized that enums/flags in C++ also missed some comments. This is also fixed by this PR.

### DIFF
--- a/src/main/scala/djinni/CppGenerator.scala
+++ b/src/main/scala/djinni/CppGenerator.scala
@@ -68,6 +68,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     val underlyingType = if(e.flags) flagsType else enumType
 
     writeHppFile(ident, origin, refs.hpp, refs.hppFwds, w => {
+      writeDoc(w, doc)
       w.w(s"enum class $self : $underlyingType").bracedSemi {
         writeEnumOptionNone(w, e, idCpp.enum)
         writeEnumOptions(w, e, idCpp.enum)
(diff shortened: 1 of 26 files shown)


src/main/scala/djinni/CppGenerator.scala
```diff
@@ -68,6 +68,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
     val underlyingType = if(e.flags) flagsType else enumType
 
     writeHppFile(ident, origin, refs.hpp, refs.hppFwds, w => {
+      writeDoc(w, doc)
```
jothepro, line 71: Without this change [this comment](https://github.com/jothepro/djinni-generator/blob/cppcli_include_idl_comments/src/it/resources/expected/my_enum/cpp-headers/my_enum.hpp#L8) would be missing.

jothepro (commented):
Most of the changes are just comments being added to the integration tests to make sure copying the comments work as expected. The only relevant changes are in `CppCliGenerator.scala` & `CppGenerator.scala`

a4z (approved):
Awesome come back :-)

freitass (approved):
Much appreciated, thank you! :)

PR #84, opened 2021-09-15
move C wrapper parameters to a separate section

jothepro:
I'm not sure what the short term plans for the c wrapper are, but to highlight their (at least theoretical) independence from Python I had the idea to also give the C wrapper parameters a separate section in the docs?

What do you think about that? Is it a bit to early? May this be confusing for users trying to use the Python interface and missing out on the required C wrapper? Or is this just the way we want to go?


a4z (approved):
Excellent move into the right direction! Thanks Johannes!!

freitass (approved):
Good one!

PR #85, opened 2021-09-16
Modify test to cover a previous C++/CLI issue

freitass:
C++/CLI issue was fixed in https://github.com/cross-language-cpp/djinni-generator/pull/78.

PR #86, opened 2021-09-16
add test for #78

jothepro:
@freitass thanks for providing me with details for what has been fixed with #78 !
#85 has helped me understand what the problem is. But I don't like how the changes to the integration tests mix up two different test inputs (by `my_enum.djinni` being imported in `my_record.djinni`). 

I propose to test the aspect in `all_datatypes.djinni` with it's focus being on correct data type marshalling. I'm still not super happy about that, but I think it's a little bit cleaner?... 🤷


src/main/scala/djinni/CppCliMarshal.scala
```diff
@@ -27,7 +27,7 @@ class CppCliMarshal(spec: Spec) extends Marshal(spec) {
   def typename(ty: TypeRef, needsHandle: Boolean): String = toCppCliType(ty, None, Seq(), needsHandle)
   def typename(name: String, ty: TypeDef): String = idCs.ty(name)
 
-  override def fqTypename(tm: meta.MExpr): String = toCppCliType(tm, Some(spec.cppCliNamespace.replace(".", "::")), Seq(), needsHandle = true)
+  override def fqTypename(tm: meta.MExpr): String = toCppCliType(tm, Some(spec.cppCliNamespace), Seq(), needsHandle = true)
```
jothepro, line 30: I'd like to discuss if the replacing is required/appropriate here.
- While not explicitly stated in the docs, my understanding is that `--cppcli-namespace` should be a C++-style namespace, so the replacement is not necessary?
- If we decided to do namespace conversion to also allow C# style namespaces, I think the conversion should happen in `Main.scala` before being set on the `Spec`?

freitass: I think the replacement made sense with the original flag: `--cs-namespace`. Now I agree that it's somewhat redundant.

freitass (approved):
This makes a lot of sense to me. Thanks for taking the time for doing it!
I'll close #85 as it's no longer required.

src/main/scala/djinni/CppCliMarshal.scala
a4z: Just a question, a C++ namespace can not contain a `.` 
So if you have the replacement, wouldn't it be up to the user to either take a C# pr C++ namespace, and both would just work?

jothepro: If we do the namespace replacement in-place just here, I am not sure it will work. There may be more places that would need the namespace to be converted. 🤔 
I think I removed the C#-namespace support when converting the CLI parameters from `--cs-*` to `--cppcli-*`.
If we decided to support both namespace types, we'd need to check that, I am not sure about it.

a4z: Thanks for info.
If it's complicated, do not do it in this context.
If it's needed, we make a new ticket and take that extra

PR #90, opened 2021-09-22
fix extern djinni flags not being marshalled properly for JNI

delaitre-manfrotto:
In djinni, `flags` are declared as EnumSet in java.

Djinni generates files named `jni/NativeXXX.hpp` which are responsible
of resolving the library symbols by looking for function signatures and
map them to Java functions.

When a djinni type `flags` is declared in one module, and used in
another one (via `@extern` and yaml files), a function `foo` taking an
extern flags `my_flags` type as parameter would try to be resolved with:

const jmethodID method_statusesChanged{
  ::djinni::jniGetMethodID(clazz.get(), "foo", "(Lnz/co/abc/MyFlags;)V")};
 
This will thus fail to be resolved as the Java type for flags is
`EnumSet`, not the actual underlying enum type.

This commit fixes this by using an `EnumSet` properly for extern flags
resulting in the following:

const jmethodID method_statusesChanged{
  ::djinni::jniGetMethodID(clazz.get(), "foo", "(Ljava/util/EnumSet;)V")};

### DIFF
--- a/src/main/scala/djinni/JNIMarshal.scala
+++ b/src/main/scala/djinni/JNIMarshal.scala
@@ -75,7 +75,10 @@ class JNIMarshal(spec: Spec) extends Marshal(spec) {
       case MSet => "Ljava/util/HashSet;"
       case MMap => "Ljava/util/HashMap;"
     }
-    case e: MExtern => e.jni.typeSignature.get
+    case e: MExtern => e.body match {
+      case e: Enum if e.flags => "Ljava/util/EnumSet;"
+      case _ => e.jni.typeSignature.get
+    }
     case MParam(_) => "Ljava/lang/Object;"
     case d: MDef => d.body match {
       case e: Enum if e.flags => "Ljava/util/EnumSet;"


src/main/scala/djinni/JNIMarshal.scala
```diff
@@ -75,7 +75,10 @@ class JNIMarshal(spec: Spec) extends Marshal(spec) {
       case MSet => "Ljava/util/HashSet;"
       case MMap => "Ljava/util/HashMap;"
     }
-    case e: MExtern => e.jni.typeSignature.get
+    case e: MExtern => e.body match {
+      case e: Enum if e.flags => "Ljava/util/EnumSet;"
+      case _ => e.jni.typeSignature
```
a4z, line 80:
```suggestion
      case _ => e.jni.typeSignature.get
```

a4z (changes requested):
Looks basically good to me, but the fix I suggest need to be applied.

And I would like to have a second opinion.

I am a bit concerned that we do not have any test case for this part, that not just checks the generated text is expected, but also ensures it can be compiled. Ideas to fix that are very welcome

src/main/scala/djinni/JNIMarshal.scala
delaitre-manfrotto: I've updated the commit. Didn't realize this changed compared to the dropbox codebase (which does not contain the `.get`)

delaitre-manfrotto:
I totally agree regarding the tests however I have basically 0 knowledge of Scala and it already took forever for this simple fix (which is working fine for me now, but I am still using the dropbox repo rather than this one; I just thought I might as well send a fix for you guys).

src/main/scala/djinni/JNIMarshal.scala
a4z: Thanks!

a4z (approved):
For having no idea about Scala you did an awesome job, @delaitre-manfrotto !
And no worries about the tests, we will do that as we have time and possibility to do so. 
It was more a note to our self.

Thank you for committing this fix!
You should consider switching your version to this one, since here is where development happens, and this is also the place where you can get help if you encounter additional situations that needs some improvements.

delaitre-manfrotto:
yeah switching is going to happen at some point ; it's just our cmake build system it quite tied to how the dropbox repo is right now so this needs some love (it was all done before the dropbox repo had any kind of cmake support)

jothepro (approved):
Thanks for your contribution, @delaitre-manfrotto !

PR #95, opened 2021-09-28
Allow externs to use an absolute path in addition to relative paths.

ngmeyer:
@extern should allow for absolute paths in addition to relative paths.  Some example usages that this patch permits to become usable:

```
@extern "path/directory/somewhere/else/extern_type.yaml"

# An interface that uses this type
my_interface = interface +c {
  # Do Something with that extern type.
  const do_something(extern_type: extern): bool;
}
```

### DIFF
--- a/src/main/scala/djinni/parser.scala
+++ b/src/main/scala/djinni/parser.scala
@@ -51,8 +51,11 @@ private object IdlParser extends RegexParsers {
     var file: Option[File] = None
 
     val path = includePaths.find(path => {
-      val relPath = if (path.isEmpty) fileStack.top.getParent() else path
-      val tmp = new File(relPath, fileName)
+      var relPath = if (path.isEmpty) fileStack.top.getParent() else path
+      if (relPath.endsWith(fileName)) {
+        relPath = relPath.substring(0, relPath.length - fileName.length)
+      }
+      val tmp = if (relPath.isEmpty()) new File(fileName) else new File(relPath, fileName)
       val exists = tmp.exists
       if (exists) file = Some(tmp)
       exists


a4z:
Thank you for your suggestion, unfortunately I have to say that I do not think we should add that.

Explanation:
* The notion of absolute 'include' (extern) paths does not exist, think about the C++ compiler, python paths, whatever
* I do not see a use-case since there is the `--idl-include-path` option for the generator that allows to include files not relative to the current IDL file in a much better way than via absolute paths

I hope you understand that I close this PR.

If you still think there is something missing or that I understood something wrong, please re-open that PR, or open an [issue](https://github.com/cross-language-cpp/djinni-generator/issues) or create a topic in the [discussions](https://github.com/cross-language-cpp/djinni-generator/discussions) forum to talk about the issue.

jothepro:
@ngmeyer While I agree with @a4z that `--idl-include-path` should probably be sufficient to solve your problem, I've got a few questions about what you want to achieve, just to be sure I understand the problem: 
- It seems like the path in the example you gave is a relative path? Did you instead mean sth like this:
   ```
   @extern "/absolute/path/directory/somewhere/else/extern_type.yaml"
   ```
- If that's what you want to achieve, I wonder how you would maintain cross-platform compatibility? I'm thinking of Windows requiring a drive letter at the beginning of an absolute path. That would mean that your IDL-file would only work on Windows or Linux/macOS?
- I have a hard time understanding what your code-change does. How would this be used? It seems like you expect the absolute path to be passed as `--idl-include-path`-parameter and if the path given in the IDL matches the absolute path given to `--idl-include-path`, you would then assume that it is an absolute path? Pls correct me if I'm wrong, I'm not sure I got this right... 🤔

PR #96, opened 2021-09-28
Restore support for --objc-base-lib-include-prefix.

ngmeyer:
--objc-base-lib-include-prefix was available in dropbox/djinni to allow for base-library files to be placed in different locations than default CMake install roots.  This is necessary for in source builds.

### DIFF
--- a/src/main/scala/djinni/Main.scala
+++ b/src/main/scala/djinni/Main.scala
@@ -82,6 +82,7 @@ object Main {
     var objcppIncludeObjcPrefixOptional: Option[String] = None
     var objcFileIdentStyleOptional: Option[IdentConverter] = None
     var objcppNamespace: String = "djinni_generated"
+    var objcBaseLibIncludePrefix: String = ""
     var cppCliOutFolder: Option[File] = None
     var cppCliIdentStyle = IdentStyle.csDefault
     var cppCliNamespace: String = ""
@@ -220,6 +221,8 @@ object Main {
         .text("The prefix path for #import of the extended record Objective-C header (.h) files")
       opt[String]("objcpp-namespace").valueName("<prefix>").foreach(objcppNamespace = _)
         .text("The namespace name to use for generated Objective-C++ classes.")
+      opt[String]("objc-base-lib-include-prefix").valueName("<include-prefix>").foreach(objcBaseLibIncludePrefix = _)
+        .text("The Objective-C++ base library's include path, relative to the Objective-C++ classes. (default: \"djinni/objc/\")")
       note("\nPython")
       opt[File]("py-out").valueName("<out-folder>").foreach(x => pyOutFolder = Some(x))
         .text("The output folder for Python files (Generator disabled if unspecified).")
@@ -451,6 +454,7 @@ object Main {
       objcppIncludeCppPrefix,
       objcppIncludeObjcPrefix,
       objcppNamespace,
+      objcBaseLibIncludePrefix,
       objcSwiftBridgingHeaderWriter,
       cppCliOutFolder,
       cppCliIdentStyle,
--- a/src/main/scala/djinni/ObjcppMarshal.scala
+++ b/src/main/scala/djinni/ObjcppMarshal.scala
@@ -5,7 +5,7 @@ import djinni.generatorTools._
 import djinni.meta._
 
 class ObjcppMarshal(spec: Spec) extends Marshal(spec) {
-  final val objcBaseLibIncludePrefix = "djinni/objc/"
+  final val objcBaseLibIncludePrefix = spec.objcBaseLibIncludePrefix + "djinni/objc/"
 
   private val cppMarshal = new CppMarshal(spec)
   private val objcMarshal = new ObjcMarshal(spec)
--- a/src/main/scala/djinni/generator.scala
+++ b/src/main/scala/djinni/generator.scala
@@ -79,6 +79,7 @@ package object generatorTools {
                    objcppIncludeCppPrefix: String,
                    objcppIncludeObjcPrefix: String,
                    objcppNamespace: String,
+                   objcBaseLibIncludePrefix: String,
                    objcSwiftBridgingHeaderWriter: Option[Writer],
                    cppCliOutFolder: Option[File],
                    cppCliIdentStyle: CppCliIdentStyle,


a4z:
@ngmeyer , I appreciate you enthusiasm and I am very thankful about your drive to contribute, 
but I would wish you discuss things with us upfront since this might avoid some confusion, and maybe even frustration.

We will not restore this flag, it is not needed. The include path in the support lib has been adopted so only the folder where the djinni root include dir is needs to be found by the compiler. 
See for example the generated include path here: https://github.com/cross-language-cpp/djinni-generator/blob/5644f72ab05cd722f05a398fcde18a1f0df300dc/src/it/resources/expected/my_record/objcpp/ITMyRecord%2BPrivate.mm#L5

The dropbox support lib had different include paths.

When we remove it we were aware that this might be a surprise for some, and maybe even a breaking change, but it is not breaking in the sense that  something does not work anymore.
In fact, this fixes something that was broken, namely the 'relative' includes for each language.
Now you just need 1 include, and if the support lib is installed on a default location, none, since it will be picked up.

If you have a cmake project, I would recommend using the support lib [via conan](https://conan.io/center/djinni-support-lib), than includes will just fit.
And if you seriously do in source build with cmake, I would consider changing that since no one, not even the cmake maintainer, think this is a good idea (even if I have a hard time to see the relation from the in source build to the location of the support lib headers)

ngmeyer:
We @Google have a set of patches that we have maintained (some for many years) as we brought Djinni into our source that are required for it to function in our build system.  Part of the effort in upgrading to this repository is to attempt to move some of those patches into PRs to reduce our customizations of Djinni. This is one of the newer ones that was required after moving from the archived dropbox/djinni repo to this newer version.  Unfortunately, in Objective-C usage we prefer absolute paths for all file references adding the include prefix to the djinni-support-lib location is beneficial for consistency. 

The prefix directories while not needed for the main usages via CMake or Conan do still have utility for other usages that should be considered.

jothepro:
Afaik when we decided to remove the parameters the assumption was that nobody would need theme anymore. I'd be open to re-add them if they serve a purpose for your usecase. @a4z correct me if I'm wrong but I think there is nothing that technically stops us from re-adding the parameter? I'm not sure if we should restore just the one that you need, or all include-prefix parameters for consistency? What's your thoughts on that? 🤔 
Just out of curiosity: It seems like you don't use the cmake target that comes with the support-lib? How do you build the support-lib? If you use another (meta) build system, would you be interested in contributing your build configuration to the support-lib as an alternative to CMake?

freitass:
> Unfortunately, in Objective-C usage we prefer absolute paths for all file references adding the include prefix to the djinni-support-lib location is beneficial for consistency.

@ngmeyer please help me understand how the `--objc-base-lib-include-prefix` flag helps with absolute paths. Even if you're consuming the support-lib from source you still have all the headers in the proper structure so, even though we recommend CMake or Conan, they are not actually needed. (I'm even wondering if your explanation actually applies to https://github.com/cross-language-cpp/djinni-generator/pull/95.)

a4z:
It seems we are talking about adding support for a very unique patch set for a very unique build system.

Because I think I can make a bazel example with the current generator/support-lib pair that will work and also find the include in the first lookup. 
Would such an example help?
Maybe that would be a better solution,  providing a bazel example ... (I know, chrome does not use bazel)  

I do not say we will not support you, but I want to carefully think about that.
We have a plan with the development, and you ask us to undo an important part.

The problem is not only just not having the command line flag, my problem with them is that  this is code that is not automated tested. 

So we constantly life in fear that will break, and since this is something no one should ever use anyway, imho, we removed that option.

So we are not talking about just adding back that option, but also adding tests. (Is that called but a ring on it, in the google book of engineering ? I have forgotten ....) 
But you get the point. This project is driven from people in their spare time, adding tests for this command line arg is expensive for us.

So before we open a box and add a snowflake solution, like it would be with just objective-c flags, we need to carefully think about the impacts, and if there are no standards way to solve the situation instead of adding just an other layer of code.
Because I think there is a way that is over time better.

a4z (changes requested):
I reopen the PR since we are not done with the discussion, but do not merge.

ngmeyer:
@freitass - I used the term absolute path instead of a path relative to the build root. Here is a more complete example of the usage of Bazel.  First for background this document discusses the output directories: https://docs.bazel.build/versions/main/output_directories.html

@jothepro - No we do not use CMake or Conan.  Everything is using Bazel.  So we have our own Build scripts that have replaced CMake.  There is no `install` concept in bazel so most directory references need to be build-root relative paths.  I am happy to share the Bazel build files that we use to compile the support-lib.

@a4z - For more background on the usage with Bazel.

First the djinni libraries are located in the third_party subdirectory, the following is how it would be referred to in a standard header import/include. 
```
#import "third_party/djinni/djinni_support_lib/djinni/jni/djinni/objc/DJIMarshal+Private.h" 
```
The generated source then is created in a specific directory relative to the build root, but the path is then appended wherever in the directory setup for the project setup.  So as an example for a new project it might be something like:

```
"bazel-genfiles/pkg/includedir/my-interface/MyRecord.h"
```

With Bazel the [objc_library](https://docs.bazel.build/versions/main/be/objective-c.html#objc_library) rule does have an ability to specify an `includes` which could be used, however this is a relative path to a few select build roots:

>The paths are interpreted relative to the package directory, and the genfiles and bin roots (e.g. bazel-genfiles/pkg/includedir and bazel-out/pkg/includedir) are included in addition to the actual client root.

So when configuring the build for these djinni generated targets I needed to calculate the relative directory from the gen target using `../../...` which was not scalable. 

The prefix solution is preferable where all include paths can have the build root as a prefix to the include/import path.

@jothepro - For other languages the prefix would be ideal to add back as well, but I've been able to work around all of the others due to a quirk in Bazel that adds the project root directory as an include for dependencies in C++ and Java packages.  In our case the Bazel BUILD file is located at the root where all of the support headers are relative to so the compilers are able to find the files.

If I updated this PR to include all of the include prefix options that be accepted?

jothepro:
>  I am happy to share the Bazel build files that we use to compile the support-lib.

Nice! I'm looking forward to your PR in the support-lib!  I think we'll have to figure out if we really want to add and maintain an additional build configuration, but maybe it will be useful to somebody else! At least having a bazel branch to begin with would be nice, I think. 😊

> If I updated this PR to include all of the include prefix options that be accepted?

I can not and don't want to make that decision on my own, but based on the information that you gave I think we should consider undoing the removal of all `--<lang>-base-lib-include-prefix` options:
- I don't know about bazel and don't want to invest too much time into understanding how it functions. If you're saying that the options are needed with your bazel build, I'll trust you on that.
- It sounds like the removal of the options caused you some trouble already, and abusing "quirks" in your build system sounds like this may haunt you in the future? On the other hand I see little to no cost in us maintaining those options. They don't come with a lot of complexity, writing proper tests for them like proposed by @a4z wouldn't be too hard IMO.
- I feel like having a lengthy discussion about why you need this is not an efficient use of our time since the existence of those options didn't cause any pain in the first place and we removed them under the assumption that nobody would notice.

> Because I think I can make a bazel example with the current generator/support-lib pair that will work and also find the include in the first lookup. (@a4z )

If you found a better solution not involving the options that would be cool! I'm looking forward to your example and to learn sth. about bazel! @ngmeyer maybe there is a better solution for your problem indeed? But still I'd like to avoid being too strict about the removal because we didn't have to remove the options for a technical reason.

a4z:
I find this is a very interesting case, and there is a lot to learn for us from this situation. (e.g, to start with, #97 :-) ) 

We will of course solve it so everyone is happy and we have technical good solution, but I want to find a good way doing so.
I see a good chance to improve what we have in several aspects, so let allow us to invest a little bit more time into exploring the problem space and finding some solution.


To me it seems, in case of a similar uscase in C++ in bazel a  `cc_library` for 3rd party lib and using the `include` together with the `strip_include_prefix` would be fine.

And for generated files it should be relatively similar, add them to a cc_library, and thread them with include and strip_include_prefix, and it should work,

If the `objc_library` is missing the `strip_include_prefix`, maybe that should be fixed in bazel, because what is with libraries that would come without generator that allows adoption of the include? 

But anyhow, 
I think we should **add bazel support to the support-lib** and **provide a genrule for the generator**.
In fact, I already thought about that. 

So it becomes a ready to use `http_archive` that can be built from within bazel without user needing to add a custom BUILD.bzl for the download.

And than a small sample program that shows how that could be used. (maybe using the existing example_cc repo)
That would allow us to reproduce the use case

@ngmeyer , it would be awesome if we could do that together, since I am far far away from being a bazel expert.

I do not think we need to build the generator with bazel, I guess the BUILD file for the generator just exports the binary as a tool, and maybe we can include a generic genrule.
And it would be super nice if the support lib could have a bazel build file included so it will be just a ready to use dependency.

I am aware that this will not be a super fast available solution, given the litte spare time everyone of us has for that task, and given the fact that it is maybe just me with some (very) basic bazel knowledge in the current maintainer team.

Therefore, Neal, If it's urgent for you, and you can not wait until this is done, we can mean while reactivate the objc rule to work around the bazel shortcoming of stripping the include prefix for objectvC libs, mark it as deprecated, and remove it in 6 month (just to have a motivation to implement the bazel solution)

ngmeyer:
@a4z I started a branch of the generator with the necessary bazel build files. Thanks to existing BUILD files this should not take much time to port over.  https://github.com/ngmeyer/djinni-generator/tree/bazel

a4z:
@ngmeyer , I checked the build files, some questions, 
but I added that in a PR to your branch
please see https://github.com/ngmeyer/djinni-generator/pull/1

jothepro:
Whats the benefits of providing a separate bazel build for the generator? Providing a bazel configuration for the support-lib sounds reasonable to me, it would help bazel-users to easily integrate the lib into their project. But the generator should just be a binary used by the build system, IMO? I don't see value in maintaining a separate build process there. The dependency versions in [djinni-generator/bazel](https://github.com/ngmeyer/djinni-generator/tree/bazel) are already out of sync (e.g. scopt 3.2 in bazel vs 3.7 in sbt). To me this just seems like it introduces unnecessary maintenance overhead. Without having done research on how this could work, IMO any bazel wrapper for the generator should just download the prebuilt library from gh, there should not be a need to build it again? 🤔

a4z:
yes, the generator could be a binary download,  but from a bazel POV it's also a question of hermetic builds.
Some bazel projects even bring their compiler via source, and compile that one into their sandbox.
so I see that it is something. and I have nothing against trying it. So please let us try it out, @jothepro 
About the versions in various files and the maintenance of that, we will of course have too look into that, but first let us just test if the whole thing would work and sketchup a POC.

jothepro:
I see, that makes sense. I didn't know about this approach, I'll have to do some research on the topic. 

I have two naive questions about this approach (I'm not trying to argue against it, just trying to understand):
- In the build some dependencies are pulled in from maven as binaries. Shouldn't they be built from scratch as well? Where do you draw the line of what should be built inside the sandbox?
- Would a bazel wrapper that initializes the sbt build also fulfill the requirements? E.g. by calling `sbt assembly` from bazel and take the resulting binary? This could avoid having to maintain separate build configurations.

a4z:
This  
> * In the build some dependencies are pulled in from maven as binaries. Shouldn't they be built from scratch as well? Where do you draw the line of what should be built inside the sandbox?

is a good and valid question


> * Would a bazel wrapper that initializes the sbt build also fulfill the requirements? E.g. by calling `sbt assembly` from bazel and take the resulting binary? This could avoid having to maintain separate build configurations.

In theory that could work,t, but specifying that might be ... complicated, since bazel tens to copy everything into some build foder or sandbox so it needs to know about every tool and how to run that, and that would include sbt
And I guess there are some advantages of using the bazel rules, when operating in bazel

There are of course also things I am curious about, and this is, why does bazel pull in portobuffer java for us even if we do not use that. This is, if you have no shared cache, not good (for first time user experience, or build bots)


btw, here is an other djinni bazel implementation
https://github.com/Snapchat/djinni

ngmeyer:
> > * Would a bazel wrapper that initializes the sbt build also fulfill the requirements? E.g. by calling `sbt assembly` from bazel and take the resulting binary? This could avoid having to maintain separate build configurations.
> 
> In theory that could work,t, but specifying that might be ... complicated, since bazel tens to copy everything into some build folder or sandbox so it needs to know about every tool and how to run that, and that would include sbt And I guess there are some advantages of using the bazel rules, when operating in bazel

From a bazel user perspective - I believe it simplifies the process to use supported build rules from the bazel tool rather than a role your own build script.  This discussion is a long issue with many projects where CMake, make, bazel, and the many other build tools are an option.  In the end each user will have their system they are going to use, and the question becomes whether the project wants to support it and ensure compatibility or not.  In this case this script is based upon one that has existed unchanged for 4-5 years.  Given the stability of the scala djinni generator changes to the build rule are unlikely beyond dependency updates or major refactoring. 

> 
> There are of course also things I am curious about, and this is, why does bazel pull in protobuf java for us even if we do not use that. This is, if you have no shared cache, not good (for first time user experience, or build bots)

I think the proto references in my `WORKSPACE.bazel` can be removed, those were brought over from the scala-binary documentation example, and appear to not be needed.

> 
> btw, here is an other djinni bazel implementation https://github.com/Snapchat/djinni

I am aware of this and used this as a starting place.  

This branch of djinni added direct support for protobuf messages, which we have multiple internal implementations for as well @ Google. I am interested in contributing protobuf support to the project once we have completed migration to a single protobuf implementation.

a4z:
The problem with the snapchat djinni is, it does also not work :-) 

The protobuf is pulled in even if I remove the explicit lines is very annoying since it gives bad first compile time.
This is not good when developing a POC. Most devs are not attached to a big remote cache on a fast network.
It's not needed now, so how to get rid of it. Removing lines that explicit request it seems not to be enough.

And the most important question at all: how to make the produced jar/binary usable at all, because currently it isn't

ngmeyer:
Inconceivable!  🤣 

I'll see what I can find today over in my dev branch.

a4z:
bazel run, java -jar, and running the binary, please look at the PR I added to your branch

I am on a Ubuntu 18.04 now, and the sbt build works. 

Maybe it will run on my Mac later in the evening, I guess you are on a Mac? 
We will see. 
But we need also Linux support, at least Ubunut and RHEL

a4z:
Update, the snapchat djinni works now on my Linux, (needed to remove some mac stuff to be able to build it) 
 only with `bazel run //src:djinni`
so there is hope :-)

jothepro:
@ngmeyer can you give me an update on the original topic of this discussion? Do you plan to adjust your build to work without the `--<lang>-base-lib-include-prefix` options or do you prefer to undo the removal, at least for some while?

a4z:
Since we haven't got any feedback about that issue, I close issue.
I also think this change is not required, and that it is possible to do a bazel build without that change. 
POC is WIP
Please feel free to reopen this issue if you still think it is a requirement.

PR #99, opened 2021-10-01
Add json serialization/deserialization support for c++ generator

mutagene:
Adapted from @hiennguyenle's work in https://github.com/hiennguyenle/finn.

Unlike Hien Nguyen Le's work, this PR does not introduce a JSON type which can be marshalled across language boundaries - in this PR, the JSON serialisation is strictly used to serialise and deserialise from the c++ datatypes.

This PR does not include the `json+extension.hpp` required for `std::optional` support from Hien Nguyen Le's json support as that would have to be included in the `djinni-support-lib`.


a4z:
Awesome! Thanks a lot!!

It will take me some time to look through that but it seems nice on a first glance.

One question I have is, how tight is the coupling to nlohmann/json? I was never a fan of that library since it's slow in every aspect, and the only selling point was a modern interface back then when c++11 was modern.

I think the decision we make now for the serializer lib will stick with us, so I would at least highlight this question since we are going to introduce dependencies.

The other thing is, I wonder if we could get somehow some usage example. 
Having test in expected generated text is good and needed.
But we need also to see how that actually works and see how to make functionality available for users. A hello world, because if there is no quick start that shows how to use a feature, it's like not having it

mutagene:
> One question I have is, how tight is the coupling to nlohmann/json? I was never a fan of that library since it's slow in every aspect, and the only selling point was a modern interface back then when c++11 was modern.

Ahhhh, yes, well, the coupling is pretty tight. hiennguyenle was using nlohmann's `adl_serializer` in the nlohmann namespace for the record, so I did the same with the enum - and to be honest, we're using nlohmann json in our project, so this would work out pretty well for us. I'm not sure how best to reduce this coupling while still retaining the benefits of "free" json serialization/deserialization (for consumers who are happy with nlohmann json).

We could pull the serialization/deserialization code out of the nlohmann namespace and put it in the djinni type's namespace, but we're still a bit constrained by the 
If we used the more flexible approach of putting the `to_json`/``from_json` methods in the type's namespace, we could leave the actual json implementation templated - but we'd still be pretty much constrained by how nlohmann json is doing things.

> I think the decision we make now for the serializer lib will stick with us, so I would at least highlight this question since we are going to introduce dependencies.

Fair enough.

> The other thing is, I wonder if we could get somehow some usage example. Having test in expected generated text is good and needed. But we need also to see how that actually works and see how to make functionality available for users. A hello world, because if there is no quick start that shows how to use a feature, it's like not having it

Yes, I agree - of course I tested the generated code in my own toy project - but I wasn't sure where such a project should be included in the djinni-generator repo.

paulo-coutinho:
Hi,

My suggestion is to use quotes instead of "<...>" to follow other headers pattern.

Thanks.

mutagene:
> My suggestion is to use quotes instead of "<...>" to follow other headers pattern.

Is this with reference to the `<nlohmann/json.hpp>` include?

a4z:
I have a little bit of a high workload this week and not sure how much time  I am able to invest into my favorite spare time projects.

But as a question and thought experiment:

Do you think it would be possible to have an implementations like 

`--cpp-json-serialization  <None/nlohmann>` 

and if future we might have , for example 

`--cpp-json-serialization  <None/nlohmann/rapidjson>`
 
and kind of 'hotsap' the json detail implementation at generation time

We do not need an other implementation right now, but just conceptually preparing this kind of flexibility, so we can control that option in future and don't close the door for users that might use already other json frameworks in their project.

a4z:
> > My suggestion is to use quotes instead of "<...>" to follow other headers pattern.
> 
> Is this with reference to the `<nlohmann/json.hpp>` include?

I do not think that should change since this is a 3rd party header and should be threaded as a system header

mutagene:
> Do you think it would be possible to have an implementations like
> 
> `--cpp-json-serialization <None/nlohmann>`

This makes sense to me, and isn't too onerous to do now. If more json libraries are added, it seems to me that it will be a bit ugly to continue shove them also into `CppGenerator.scala` as I have done now with the nlohmann implementation, but I'm not prepared or equipped to refactor CppGenerator now.

paulo-coutinho:
Wow! It is nice!

a4z:
Thanks for adding the flexibility option! Making it more pretty once we might have multiple json lib support is a future problem.

After a first glance, I think this looks good. 
I need just find a bit more time to play with it, but I think we will pull that in and have json soon support. Thanks a lot! 

It would be an awesome addition to have some doc with some small usage example, or any other way that can show people how to have an easy start into using that feature.

mutagene:
> @mutagene , a SwedenCpp user group colleague , @bigdavedev , shared that with me https://godbolt.org/z/1hna1G1zM Do you think that could be a log-friendly and reasonable good solution ?

Ahh, nice - yes, that's quite lovely.

mutagene:
> Thoughts?

I was thinking a bit the same when I made the last commit. On the other hand, if we don't provide an implementation at all, then projects using `date` with json serialization enabled simply won't build.

I guess one solution could be to take @paulo-coutinho's solution and make this the default. If someone wanted to override it with the `date.h` implementation, they could write their own `adl_serializer` as outlined in the documentation:
```
#include <nlohmann/json.hpp>
#include <date/date.h>

namespace nlohmann {
    template <>
    struct adl_serializer<std::chrono::system_clock::time_point>
    {
        static void to_json(json &j, const std::chrono::system_clock::time_point& tp) {
            j = date::format("%F %T %Z", tp);
        }

        static void from_json(const json &j, std::chrono::system_clock::time_point& value) {
            if (j.is_null()) {
                auto dur = std::chrono::milliseconds(0);
                value = std::chrono::time_point<std::chrono::system_clock>(dur);
            } else {
                std::istringstream json_time{j.get<std::string>()};
                std::chrono::system_clock::time_point parsed_time{};
                // Time saved in UTC, so no need to extract time zone
                json_time >> date::parse("%F %T", value);
            }
        }
    };
}
```
- although this would have to be included before `json+extension.hpp`, which is a bit of a pain 💦

a4z:
> I was thinking a bit the same when I made the last commit. 

I am super happy we have the same point of view! 

> On the other hand, if we don't provide an implementation at all, then projects using `date` with json serialization enabled simply won't build.

That could be fine , to get a compiler error, but maybe there is one more way 
 
> I guess one solution could be to take @paulo-coutinho's solution and make this the default. If someone wanted to override it with the `date.h` implementation, they could write their own `adl_serializer` as outlined in the documentation:

 Since we are in C++ land, we can always decorate something with defines to remove it 😁 


```
#include <nlohmann/json.hpp>
  
namespace nlohmann {

#ifndef DJINNI_CUSTOM_JSON_DATE
     template <>
     struct adl_serializer<std::chrono::system_clock::time_point>
     {
         /// .. some default implelemtation
     },
#endif  
```
and then this 

> -- although this would have to be included before `json+extension.hpp`, which is a bit of a pain 💦

is not a problem anymore

I will add that as review suggestions. 

I think with such adoption we should be done. Great discussion leading to awesome improvements and a real powerful library feature.  Thanks a lot !!

PR #101, opened 2021-10-05
add code formatter

paulo-coutinho:


paulo-coutinho:
I have added a hook into **.git/hook** folder, but it don't go to github. Anyone knows how to do it?

```
#!/bin/sh
sbt plugins scalafmtAll
```

a4z (changes requested):
Nice, that looks already good, thanks a lot! 

Paulo, could you please splitt that into a 2 step operation

1) Add scalafmt as a tool so it can be called via sbt (and to the docs, extra thanks for thinking about the docs!) 

This way we can play around and use it local, see how to fine tweak it.
There are some minor issues, like it reformat the license text header what should not happen if it could be avoided.
So maybe we need to find out some config options

2) Once we have the tool included, we will do an additional  PR with only the reformatted code

--
I think solving this in those 2 steps will gives a much cleaner way of getting the code in good shape.
Do you think that is a good idea?

PR #102, opened 2021-10-06
feature scala code formatter

paulocoutinhox:


a4z (approved):
Awesome, thanks a lot!

README.md
```diff
@@ -70,6 +70,20 @@ Add the generator as a build requirement in `conanfile.txt`:
...
+
+To format all files execute `sbt plugins scalafmtAll`.
+
+The version of scala format tool can be changed in the file `.scalafmt.conf` and the version of scala plugin can be changed in the file `project/plugins.sbt`.
+
+To change the code formatter configurations, put all configurations options inside `.scalafmt.conf`. See all configutations here: https://scalameta.org/scalafmt/docs/configuration.html. 
+
+You can check if formatter is working with command `sbt scalafmtCheck `.
+
+All available tasks of the code formatter plugin can be checked here: https://scalameta.org/scalafmt/docs/installation.html#task-keys.
```
jothepro, line 85:
```suggestion
### Code Formatting

All **scala**- and **sbt**-files need be formatted with the [scala-fmt](https://scalameta.org/scalafmt/) plugin.

To format all files, execute `sbt plugins scalafmtAll`.

The version of the scala format tool can be changed in the file `.scalafmt.conf` and the version of the scala plugin can be changed in the file `project/plugins.sbt`.

To change the code formatter configurations, put all configuration options inside `.scalafmt.conf`. See all configuration options [here](https://scalameta.org/scalafmt/docs/configuration.html). 

You can check if the formatter is working with the command `sbt scalafmtCheck `.

All available tasks of the code formatter plugin can be checked [here](https://scalameta.org/scalafmt/docs/installation.html#task-keys).
```

jothepro (commented):
Thanks @paulo-coutinho for your PR!

- Do you have an idea on how the formatting could be enforced? Maybe it should be checked in the pipeline?
 My personal experience is that a formatter that is not strictly enforced actually makes maintenance harder, because random reformats happen to bloat PRs: If one person forgets to format properly, it will mess up the next PR by a person that did format properly and thereby formatted the other persons changes as well.

- At the risk of starting a religious discussion about formatting preferences I'd like to discuss 2 formatting decisions upfront:

  - I don't enjoy what the formatter does to block comments:
     ```scala
       /** Gererate the definition of Objc constants.
         */
       def generateObjcConstants(
     ```
     The same problem exists for the license headers, what the formatter does to them just hurts my eyes... 😢 
  - The maximum line length seems to be 100 characters, which seems odd. Isn't 120 a more widely used default?

a4z:
When this is merged, and we found some format config that is suitable, we can add the formatter to the PR checks and if a file need to be reformatted, the CI will fail.
scalafmt, like all other formatter of quality, can do that, telling you if it was applied to a file or not.
This way developers will have to use it for contributions.
I think this will be the proper integration.

paulo-coutinho:
Hi,

Fixed. Now it works on terminal and on REPL.

PR #105, opened 2021-10-07
Update scopt

### LINKED ISSUE #98: Update scopt to 4.0.1
[scopt 3.7.x is only compatible with scala <= 2.13](https://mvnrepository.com/artifact/com.github.scopt/scopt)

In order to potentially be able to update to scala 3 somewhere in the future, we'll have to update scopt to the lastest major version in the first place. This involves some refactoring of the CLI parsing.

While doing so, it may be interesting to discover if there is a way to deprecate cli options in the new scopt version. (#97 )

paulocoutinhox:
fixes #98


paulo-coutinho:
> That was... way easier than expected? 🙈 Thank you!

Was easy. I only need read full docs and samples huahuaahuahuhauhuaahu.

PR #106, opened 2021-10-07
Format all the things 

### LINKED ISSUE #100: Use scala code format tool (in CI)
Hi,

My suggestion is to use a scala format tool, like this:

- https://scalameta.org/scalafmt/
- https://github.com/scalameta/scalafmt

It will make the code more clear and easy to do maintenance.

Thanks.

a4z:
All the things? No, 
unfortunately, I am unable to format the tests ....
and also check the format for the tests

but beside that,  simply apply the default format for all sources in src/main, and add a gh action that checks that those files are in shape.

If there are any ideas how to apply scalafmt  + checks  to the PR , please do so, to close #100

### DIFF
--- a/src/main/scala/djinni/Main.scala
+++ b/src/main/scala/djinni/Main.scala
@@ -125,8 +125,8 @@ object Main {
             }
           )
       }
-      
-      override def showUsageOnError = false
+
+      override def showUsageOnError = Some(false)
       head(
         "djinni generator version",
         Main.getClass.getPackage.getImplementationVersion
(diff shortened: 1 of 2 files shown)


paulo-coutinho:
Hi,

I think that the problem can be the name of project "it".

If you rename it to "test", you can run `sbt test:scalafmt` and it will work.

paulo-coutinho:
I found the problem. Tests are in wrong path. To keep it inside "it" folder and format works, you need add line `scalaSource in Test := baseDirectory.value / "src/it/scala",` after `Defaults.itSettings,` inside file `build.sbt`.

But i suggest do this to keep the default pattern:
https://github.com/cross-language-cpp/djinni-generator/issues/107

a4z:
Thanks for preparing all the changes, Johannes!

PR #111, opened 2021-10-13
Update toolchain

a4z:
* scala version to most recent 2.13.6
* sbt to 1.5.5
* 3rd party dependencies to latest available version

Adopt syntax / code
Adopt some expected test output to have generated lines in defined order

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -54,7 +54,7 @@ sbt "run --help"  # show all options
 sbt IntegrationTest/test
 ```
 
-For testing, the djinni target needs to exist, so run `sbt assembly` before.
+For testing, the djinni target needs to exist, so run `sbt assembly` before. This is also required when you do development on the generator. Always run `sbt assembly` before doing integration testing!
 
 ### Packaging
 


build.sbt
```diff
@@ -1,21 +1,22 @@
...
 
 lazy val djinni = (project in file("."))
   .configs(IntegrationTest)
   .settings(
     name := "djinni",
     Defaults.itSettings,
-    libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.5" % "it",
-    libraryDependencies += "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.2",
-    libraryDependencies += "org.yaml" % "snakeyaml" % "1.26",
+    libraryDependencies += "org.scalactic" %% "scalactic" % "3.2.10",
```
jothepro, line 11: @a4z do you remember why scalatic was added as a dependency in this PR? I cannot find any imports that make use of it. :/

a4z: I plan(ned) to use it, but haven't come that far (so far) 
there are some nice features for testing in it

PR #113, opened 2021-10-18
Support objc generics from extern types.

eakoli:
This is based on the generics population that already existed with the java generator

Currently the default is "false" to preserve backwards compatability


a4z:
This PR fails due to the code format check, on reason to worry.

You can simply format all the files by running `sbt scalafmtAll` 
It's a good idea always doing so before a commit or before pushing changes up to a PR.

Please see https://github.com/cross-language-cpp/djinni-generator/blob/main/docs/developer-guide.md#code-formatting for more information.

eakoli:
@a4z @jothepro I took a stab at adding a test, which also exposed a bug I had, so it should all work now =)

jothepro (commented):
Wow, Objective-C sometimes feels so alien to me... I didn't know that NSArray originally was untyped and Generics have [been introduced as a compile-time check afterwards](https://microsoft.github.io/objc-guide/Syntax/Generics.html)... 😮 

I've got a few questions to make sure I understand whats going on:
- The interface defined in `custom_array.yaml` seems kind of pointless, as it has the same types like the `array` type built into djinni. My understanding is that normally you would for example want to implement a custom array-ish type in C++ that would then be exposed as `NSArray` to Objective-C.  I guess you didn't care about this because for the test it doesn't matter what the type in C++ is? Do I get that right?
- Just out of curiosity: My perception is that you want to use this feature to expose a custom array-ish data structure from C++ to Objective-C (or vice versa?). Why don't you subclass `std::array` in C++ and implement your own logic there? This way I think you would not need to implement a custom array type?  Is my perception right or does the test just use the array as an example and you or doing something completely different with the generics?

src/it/resources/custom_array.yaml
```diff
@@ -0,0 +1,38 @@
+# This provides outcome<sucess_type,failure_type>
```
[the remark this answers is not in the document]
eakoli: Oops, I simplified the test from the outcome type we have.

eakoli:
The array was just the simplest way to show the example.

Ways that we will be using it is to expose an ordered maps and sets ( where iteration is based on the order items were inserted) which need to map to custom NSDictionary/java.util.Map and NSSet/java.util.Set implementations.

We could build a custom type w/o this but on Objc we would not have the strung typing exposed (but would on the java side).

Another use would be to expose an outcome type, snapchats for has this, but it has to be implemented in the generator. With this PR those kinds of types can be implemented purely as external types.

a4z:
> .... With this PR those kinds of types can be implemented purely as external types.
This sounds very interesting

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -412,6 +412,33 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
     }
 
+    it(
+      "should be able to include objc generic type information for extern types"
+    ) {
+      Given(
+        "an IDL-file that uses a parameterized extern type that enables Objective-C option 'generics: true'"
+      )
+      val idlFile = "my_record_objc_generics"
+      When("generatinc the Objective-C gluecode")
```
jothepro, line 422: Sry for being picky 🙈 

```suggestion
        "an IDL-file that uses a parameterized extern type that enables the Objective-C option 'generics: true'"
      )
      val idlFile = "my_record_objc_generics"
      When("generating the Objective-C gluecode")
```

jothepro (approved):
Thanks! Apart from some typos this looks good to me. 👍  Unfortunately I don't have the capacity to test this feature myself and we don't have the possibility to write a compilation test for it here in the generator (yet). But I think we should go for it if it works for you! 

@a4z do you think it makes sense to add a test for this in the support-lib? Afaik the long-term goal is to somehow execute support-lib tests in the generator pipeline as well, so it would be nice to have a compiling/running test case for this feature in the other repository for now?

a4z:
yes, I am all for getting into a testing state where we do not just compare generated text, but also have something executing. Optimally not just in a series of asserts,  but also in a way that it can be used as documentation, but this is an extra story.
To start with, adding something into support lib is a good begin, but will remind us on that https://github.com/cross-language-cpp/djinni-generator/discussions/94

--

btw, does this PR mean, that at least for  Objective-C and Java , this, https://github.com/cross-language-cpp/djinni-generator/discussions/92, is actually implemented ?

eakoli:
@a4z Yes this extends the existing "generics" support to Objective-C and would partially (depending if on what languages they needed it for) address https://github.com/cross-language-cpp/djinni-generator/discussions/92.

I have the working changes to extend it to C#, let me put those up as well

a4z:
do you want to make that in this PR or in an extra one after this is merged?

a4z:
we probably should merge that one first, and then have the C# thing extra
can you press the apply button for the type fix, the we merge this, and the C# becomes some extra PR ?

a4z (approved):

eakoli:
Didn't see your note.

I just pushed the C# changes as well.

But I reworked the test to to show the full intention a little better. It now modeled after the vendor/date example in the support-lib, showing that you can fully support "list" as an external type

src/it/resources/expected/extern_generics/cppcli/MyRecord.cpp
```diff
@@ -0,0 +1,60 @@
+// AUTOGENERATED FILE - DO NOT MODIFY!
+// This file was generated by Djinni from extern_generics.djinni
+
+#include "MyRecord.hpp"  // my header
```
a4z, line 4: the `// my header` comment is generated and expected?

a4z (changes requested):
What has happend to the tests? It seems they are not part of the PR anymroe?

src/it/resources/expected/extern_generics/cppcli/MyRecord.cpp
eakoli: hmm that's what the generator produces, you can see it in all the other generated cppcli sources in expected.

a4z: oh, indeed, its in the code 
https://github.com/cross-language-cpp/djinni-generator/blob/4649bd1ed78dec54a81be0ee563cbc9b02be6931/src/main/scala/djinni/generator.scala#L611
lol 
so lets leaf that (for now)

a4z (approved):

src/main/scala/djinni/YamlGenerator.scala
```diff
@@ -154,6 +154,7 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
     "header" -> QuotedString(objcMarshal.include(td.ident)),
     "boxed" -> QuotedString(objcMarshal.boxedTypename(td)),
     "pointer" -> objcMarshal.isPointer(td),
+    "generic" -> false,
```
jothepro, line 157: Is there a reason why the Objective-C part has a default value for the key "generic" defined here, while the C++/CLI part doesn't (`private def cs(td: TypeDecl)`)?

jothepro (commented):
Cool! Given the C++/CLI changes in this PR I've requested a PR from @freitass because he is the author of the C++/CLI bindings.

src/main/scala/djinni/YamlGenerator.scala
eakoli: No, the change was made such that the parameter is optional. The Java code is doing the same thing as here, so I mimicked it in the objc, just forgot to do the same with the CLI part.

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -412,6 +412,69 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+        cppOmitDefaultRecordCtor = true
+      )
+
+      djinni(cmd)
+
+      Then(
+        "the generic type parameters should be included in the C++/CLI type declarations"
+      )
+      assertFileContentEquals(idlFile, CPPCLI, cppcliFilenames)
+    }
```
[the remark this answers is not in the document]
eakoli: I was thinking the same thing, just wasn't sure what the best approach was, as the failure output when there are multiple "then" statement with checks isn;t really helpful, hard to actually tell what failed, so I had made it as multiples following one of the other tests that was split into multiple parts.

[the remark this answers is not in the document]
jothepro: Makes sense. Maybe one could combine the then-statements into a single one that is more broadly describing what's tested, but it's fine how it is now! 👍

src/main/scala/djinni/YamlGenerator.scala
a4z: Can you please give me an update on this? 
Because from this conversation it is not totally clear to me if we shall wait with merging this PR until we have one more change, or if it is already in it's final state and the thing you discuss here is not important.

eakoli: I pushed the corresponding change for c++/cli

a4z: Ok, thanks for the info! I guess that means it's ok to merge.

a4z:
Great that this is now in main , thanks a lot @eakoli for the contribution and our patience!

delaitre-manfrotto:
@eakoli : sorry if I did not quite understand this PR/issue ; I am the guy who asked for #92 and I am still very much interested in getting some generic support for djinni, but I am not 100% how far this PR goes with respect to what I ideally need as a user :)

So, let's say I have a the following C++ type: `template <typename T> MyType {...};`
Is this PR able to make this type available to java/objc? If so, what do I need to do to expose it?
Does it mean in my djinni files I can then use `MyType<i32>` or `MyType<AnotherDjinniType>` ?

If so, you totally made my day!

eakoli:
@delaitre-manfrotto I answered in the referenced discussion

Issue #117, opened 2021-11-01
Objective-C generated objects should conform to NSObject protocol


ngmeyer:
See best practices document from Apple here: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithProtocols/WorkingwithProtocols.html

As an example not conforming to NSObject protocol results in build errors if we attempt to get the class from the protocol e.g.:

for (id<ABCGeneratedType> thing in things) {
  NSString *className = NSStringFromClass([thing class]);  // build error
  ...
}

a4z:
Since the PR was merged, we can close that issue as solved

PR #118, opened 2021-11-01
Update the Obj-C generator to have generated protocols conform to NSObject

ngmeyer:
All Obj-C protocols will now conform to NSObject.

### DIFF
--- a/src/it/resources/expected/my_client_interface/objc-headers/ITMyClientInterface.h
+++ b/src/it/resources/expected/my_client_interface/objc-headers/ITMyClientInterface.h
@@ -4,7 +4,7 @@
 #import <Foundation/Foundation.h>
 
 
-@protocol ITMyClientInterface
+@protocol ITMyClientInterface <NSObject>
 
 - (BOOL)logString:(nonnull NSString *)str;
 
--- a/src/main/scala/djinni/ObjcGenerator.scala
+++ b/src/main/scala/djinni/ObjcGenerator.scala
@@ -137,7 +137,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
         }
         w.wl
         writeDoc(w, doc)
-        if (i.ext.objc) w.wl(s"@protocol $self")
+        if (i.ext.objc) w.wl(s"@protocol $self <NSObject>")
         else w.wl(s"@interface $self : NSObject")
         for (m <- i.methods) {
           w.wl


ngmeyer:
fixes #117

a4z:
This is the forward declaration, so that should be fine as it is
https://github.com/cross-language-cpp/djinni-generator/blob/09b420537183e0f941ae55afbf15d72c3f8b76a1/src/main/scala/djinni/ObjcppGenerator.scala#L147

but I  have a hard time remembering what that does
https://github.com/cross-language-cpp/djinni-generator/blob/09b420537183e0f941ae55afbf15d72c3f8b76a1/src/main/scala/djinni/ObjcMarshal.scala#L78
if this is triggered by some tests (not sure (note to myself, add code coverage)) and if that can stay as it is or not.
Any insights on that, @artwyman , do you maybe remember if that is covered by some test?

freitass:
> ...
> but I have a hard time remembering what that does
> 
> https://github.com/cross-language-cpp/djinni-generator/blob/09b420537183e0f941ae55afbf15d72c3f8b76a1/src/main/scala/djinni/ObjcMarshal.scala#L78
> 
> 
> if this is triggered by some tests (not sure (note to myself, add code coverage)) and if that can stay as it is or not.
> Any insights on that, @artwyman , do you maybe remember if that is covered by some test?

Here's a small example that triggers that line:
```djinni
custom_interface = interface +o {
  test();
}

other_interface = interface +o {
  custom(): custom_interface;
}
```

which results in:

```objc
// This file was generated by Djinni from using_custom_datatypes.djinni

#import <Foundation/Foundation.h>
@protocol ITCustomInterface;


@protocol ITOtherInterface

- (nullable id<ITCustomInterface>)custom;

@end
```

To be honest, I don't know if it needs to be changed for the scope of this PR.

artwyman:
I'm too rusty on ObjC to know whether forward declarations of protocols should declare their inheritance relationship, or if that could cause multiple-definition errors.  I'd guess that any code which actually tries call methods on an interface object will want to include its full definition anyway, so the point may be moot.  There could be cases where there would be a difference, though.

If you wanted to test this to see what works better, I'd suggest creating a test case corresponding to the use case which motivated this PR.  I.e. write some code which calls [object class] on an object using a Djinni protocol type.  Then see if you can do it successfully with only a forward-decl or if it requires the full definition of the protocol.  Then see if adding the protocol inheritance to the forward-decl changes the behavior.

a4z:
Thanks all together, seems that is indeed all it needs.

Yes, we need more testing, also code that is not just checked against other code, but actually executed.
We are working on that, but that takes time ... 

I think we pull it in, and in case something is missing, some one will tell us anyway ...

a4z (approved):

jothepro (approved):
From what I've read in Slack this change seems to have no known side effects. I'm fine with it 👍

eakoli:
@a4z This change forces all swift objects that implement a protocol to be derived from NSObject, instead if just being able to implement the protocol.

a4z:
Hi @eakoli , can you please provide an example that highlights the problem. And, it you think this is a problem, open a ticket so this can be discussed more publicly. Here, at a closed PR it will not be very visible. Thanks!

PR #119, opened 2021-11-18
Update scala-fmt to version 2.4.4

paulo-coutinho:


build.sbt
```diff
@@ -19,4 +19,3 @@ lazy val djinni = (project in file("."))
     assembly / assemblyOption := (assembly / assemblyOption).value.copy(prependShellScript = Some(defaultUniversalScript(shebang = false))),
     assembly / test := {}
   )
-  inConfig(IntegrationTest)(org.scalafmt.sbt.ScalafmtPlugin.scalafmtConfigSettings)
```
a4z, line 22: documentation says this is required
https://scalameta.org/scalafmt/docs/installation.html#enable-integrationtest
to also format the integration tests, and not only the source code

a4z (changes requested):
Please update the PR title, this is not a fix but a version update

build.sbt
paulo-coutinho: It is not necessary, because i open an issue. Maybe the docs are not updated.
https://github.com/scalameta/sbt-scalafmt/issues/152

jothepro: I can confirm that. Thanks, @paulo-coutinho !

jothepro (approved):

build.sbt
a4z: indeed

a4z (approved):
Thanks!

PR #125, opened 2021-12-14
better fix for JNI enum flags in yaml output

delaitre-manfrotto:
@finalpatch made a better fix in the Snapchat/djinni repo than what
was made in efabbc720591ba685e0c48c3333b31e575c7c923.

This commit uses the better solution.

@finalpatch commit is https://github.com/Snapchat/djinni/pull/33/commits/53f5a34cf03d5cef211257121ab67a98be9c4993

### DIFF
--- a/src/main/scala/djinni/JNIMarshal.scala
+++ b/src/main/scala/djinni/JNIMarshal.scala
@@ -10,7 +10,10 @@ class JNIMarshal(spec: Spec) extends Marshal(spec) {
   // For JNI typename() is always fully qualified and describes the mangled Java type to be used in field/method signatures
   override def typename(tm: MExpr): String = javaTypeSignature(tm)
   def typename(name: String, ty: TypeDef) =
-    s"L${undecoratedTypename(name, ty)};"
+    ty match {
+      case e: Enum if e.flags => "Ljava/util/EnumSet;"
+      case _                  => s"L${undecoratedTypename(name, ty)};"
+    }
 
   override def fqTypename(tm: MExpr): String = typename(tm)
   def fqTypename(name: String, ty: TypeDef): String = typename(name, ty)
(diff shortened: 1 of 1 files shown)


a4z:
Nice, thanks a lot!

I just wonder, what would be required to have testing to see that code also executed ?

delaitre-manfrotto:
Yeah adding a test for that would be nice but I basically have zero scala knowledge. It already too me forever to make my original fix because I needed that to work.
I am not sure how the tests are setup, but if you can have a djinni file declaring a flag type, run the generator and check if the resulting yaml file uses `typeSignature: 'L/java/util/EnumSet;'` then the proper type is used.

a4z:
Due to Christmas time I have some delay in reviewing this (checkout and do some testing at least locally), 
I am sorry for that, 
but I will try to catch up as soon as possible

a4z:
> Yeah adding a test for that would be nice but I basically have zero scala knowledge. It already took me forever to make my original fix because I needed that to work. I am not sure how the tests are setup, but if you can have a djinni file declaring a flag type, run the generator and check if the resulting yaml file uses `typeSignature: 'L/java/util/EnumSet;'` then the proper type is used.

did so , that is the resulting YAML

```
jni:
  translator: '::djinni_generated::MyFlags'
  header: '"cpplib/djinni/jni/MyFlags.hpp"'
  typename: jobject
  typeSignature: 'Ljava/util/EnumSet;'
```

without the change is was 

```
jni:
  translator: '::djinni_generated::MyFlags'
  header: '"cpplib/djinni/jni/MyFlags.hpp"'
  typename: jobject
  typeSignature: 'Ldev/xlcpp/djinni/hello/MyFlags;'
```

the generated Java code for 

```
my_flags = flags {
    flag1;
    flag2;
    flag3;
    no_flags = none;
    all_flags = all;
}

```

 is

```
// This file was generated by Djinni from storage.djinni

package dev.xlcpp.djinni.hello;

public enum MyFlags {
    FLAG1,
    FLAG2,
    FLAG3,
    ;
}
```

a4z (approved):
LGTM, but I would like a second view of @jothepro and/or @freitass since they have more know how about Java and Johannes especially about the YAML parts

jothepro:
@delaitre-manfrotto thanks for your attention to detail! @a4z thanks for testing it manually! I will try to contribute a test in the near future, but probably that's only going to happen next year...

jothepro (approved):

PR #130, opened 2022-05-04
update hardcoded versions to match the latest release

jothepro:
Every. Time. I forget to update this hardcoded version in the docs before the release. Last time I deleted the release and created a new one, but this time I realized it too late. 🙄

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Add the generator as a build requirement in `conanfile.txt`:
 
 ```text
 [build_requires]
-djinni-generator/1.3.0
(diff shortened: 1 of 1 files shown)


a4z (approved):
Thanks! 

Since this is a readme, maybe we should change it to something more generic
(there is the possibility to request version ranges in conan) 
https://docs.conan.io/en/latest/versioning/version_ranges.html
then we could have something like `djinni-generator/[>=1.3.0]`
and add some explanation .... 


will look into it

PR #131, opened 2022-05-13
Use simplified method for converting types to/from json

mutagene:
The basic recommended method for converting to/from json is described here: https://json.nlohmann.me/features/arbitrary_types/#basic-usage

```
namespace ns {
    void to_json(json& j, const person& p) {
        j = json{ {"name", p.name}, {"address", p.address}, {"age", p.age} };
    }

    void from_json(const json& j, person& p) {
        j.at("name").get_to(p.name);
        j.at("address").get_to(p.address);
        j.at("age").get_to(p.age);
    }
} // namespace ns
```

Where `ns` is the type's namespace.
 
The current implementation in djinni is using the method recommended for third-party types such as e.g. in std where we cannot place our implementation into the type's namespace.

When there are errors, however, it is much harder to understand what's wrong -and I've recently run into problems trying to convert types which have been defined in multiple libraries.

### DIFF
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
@@ -15,82 +15,73 @@
 #include <unordered_set>
 #include <vector>
 
-namespace nlohmann {
-
-template <>
-struct adl_serializer<::AllDatatypesJson>  {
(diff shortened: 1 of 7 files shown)


a4z:
Seems like an improvement to me, thank you! 
Since I am not a user of json now, I have no code to use and check that, but for me it looks ok.

a4z (approved):

jothepro (approved):

PR #132, opened 2022-05-20
Fixes an issue with nested generic types

eakoli:
This fixes an issue where if a generic external type is used as a template type its generic information would not be generated  properly for objc.

Example

```
my_record = record {
  optional_builtin: optional< list< string > >;
  optional_external: optional< extern_list< string > >;
  list_external: list< extern_list< string > >;
} deriving (eq,parcelable)
```

would end up generating optional_external as
```
NSArray * optionalExternal;
NSArray<NSArray *> * listExternal;
```
instead of
```
NSArray<NSString *> * optionalExternal;
NSArray<NSArray<NSString *> *> * optionalExternal;
```
thus loosing the nested type info.

Additionally this adds the same test coverage for the Java and C++/CLI generators


src/it/resources/list.yaml
```diff
@@ -18,9 +18,24 @@ objc:
...
+  typename: jobject
+  typeSignature: 'Ljava/util/ArrayList;'
 cs:
   translator: '::djinni::List'
   header: '"djinni/cppcli/Marshal.hpp"'
   typename: "System::Collections::Generic::List"
   reference: true
   generic: true
+
+# No C# definitions!
```
jothepro, line 41: Can you clarify what you mean? The C# definitions are right above the comment if I'm not mistaken?

jothepro (commented):
Thanks for adding the Java tests! 👏

src/it/resources/list.yaml
eakoli: yea..not sure how that got in there.....will remove it.

a4z (approved):
Thanks! This looks good.

There are 2 things I can not verify, and this is C# and yaml part, since I am not a user of both.

Therefore, before merging this one, I it would be good to see a second approval that is focused on that part.

PR #134, opened 2022-07-06
Make NSObject based protocols optional

### LINKED ISSUE #133: Recent change for all protocols to be based on NSObject add additional restriction on swift use.
This PR https://github.com/cross-language-cpp/djinni-generator/pull/118# updated the genreated @protocol for

```
callback = interface +o { 
   ...
}
```

from this

```
@protocol callback {
   ....
}
```

to this

```
@protocol callback <NSObject> {
   ....
}
```

When using djinni to integrate with objc its not much of an issue. But, when the target consumer is Swift, it adds an additional requirement that the swift objects be NSObjects, which limits what swift objects can implement the protocol.

Previously we were using generator 1.1.0 to generate an interface consumed exclusively by swift code. Once we updated to 1.3.0, we had to go change the derivation of every swift object implementing an interface from djinni to derive from NSObject. In some cases we had to rework the code as the class originally implementing the interface was derived from an object that we don't control the source for, so we had to switch to a light object that implemented the interface an delegated to the actual object that had the implementation.

I would propose that there is a way to disable the <NSObject> requirement. 

Maybe its auto-disabled when `--objc-swift-bridging-header` is defined, as the purpose of that switch is to generate an interface importable by swift.

Or a new switch `--objc-strict-protocols` that when ON enables this functionality.

a4z:
This might fix #133

* Add new bool toggle --objc-strict-protocols
* Enables / disables the <NSObject> tag after generated protocol definition
* Default of --objc-strict-protocols is true since this is current
  behaviour
* Tests are missing, might be added after this solution preview is
  accepted

### DIFF
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -38,128 +38,129 @@ djinni \
 
 ### General
 
-| Argument | Description |
-| -------- | ----------- |
-| `--help` | Print help  |
-| `--version` | Print version |
-| `--idl <in-file>` | The IDL file with the type definitions, typically with extension `.djinni`. |
+| Argument                        | Description                                                                                          |
+|---------------------------------|------------------------------------------------------------------------------------------------------|
+| `--help`                        | Print help                                                                                           |
+| `--version`                     | Print version                                                                                        |
+| `--idl <in-file>`               | The IDL file with the type definitions, typically with extension `.djinni`.                          |
 | `--idl-include-path <path> ...` | An include path to search for Djinni `@import` and `@extern` directives. Can specify multiple paths. |
 
 ### Java
 
-| Argument | Description |
-| -------- | ----------- |
-| `--java-out <out-folder>` | The output for the Java files (Generator disabled if unspecified). |
-| `--java-package ...` | The package name to use for generated Java classes. |
-| `--java-class-access-modifier <public/package>` | The access modifier to use for generated Java classes (default: `public`). |
-| `--java-cpp-exception <exception-class>` | The type for translated C++ exceptions in Java (default: `java.lang.RuntimeException` that is not checked) |
-| `--java-annotation <annotation-class>` | Java annotation (`@Foo`) to place on all generated Java classes |
-| `--java-generate-interfaces <true/false>` | Whether Java interfaces should be used instead of abstract classes where possible (default: `false`). |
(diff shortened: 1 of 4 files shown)


paulocoutinhox:
Hey guys, can you add focus to solve this, please? It will help me a lot.

a4z:
Hi @paulocoutinhox , did you have a chance to test this PR, and check if the toggle works for you? 
If yes, please report back, this will help accelerating the solution. Thanks!

paulocoutinhox:
Hi,

Don't know if im doing something wrong, but this is generating the objc files in both cases with NSObject.

I made a repository to test this:
https://github.com/paulocoutinhox/djinni-objc

Thanks.

paulocoutinhox:
Hi,

With `+o` it works.

I update my test project to store true/false values.

Only need check if with `+c` it need change,

a4z:
yes, we are talking about the protocol implementation, not the interface one

and that works, 
`interface +o {....}` idl generates a `@protocol NTVWhatever` where you can toggle the NSObject base
please note the **@protocol**

 `interface +c {....}` idl generates a `@interface NTVWhatever : NSObject` 
please note the **@interface** , and that behaviour has, I think, always been like that and it not topic of  #133/


* `interface +c {....}` will be implemented in C++ and can be called from any language. 
* `interface +j +o +p +s` will be implemented in Java, ObjC, Python and C# and can be called from C++.

we are handling the second case here, not the first one

if there are any concerns about the NSObject base for Objective-C classes, this must be an extra topic / issue, but I do not thing that not using NSObject for OC interfaces/classes is a good idea

paulocoutinhox:
Ok, nice.

paulocoutinhox:
Hi,

According to Apple docs (https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithProtocols/WorkingwithProtocols.html), only protocol can be without NSObject. Interfaces need have NSObject.

This PR is OK for me.

Thanks.

jothepro (approved):
Thx @a4z for the change!

Thx @paulocoutinhox for testing the change! I haven't tested it but it looks good to me 👍 

Would be nice if we had a test for it, maybe I can assist with that in the next weeks 🤔

a4z:
> Would be nice if we had a test for it, maybe I can assist with that in the next weeks 🤔

We do not have any exciting infrastructure for command line options test, afaik. And we have a lot command line options. Having some progress in that direction would be awesome.

PR #136, opened 2022-09-01
Make json serialisation methods static inline

mutagene:
Previous switch from `nlohmann`-namespaced `adl_serializer(...)` method to putting the `from_json`/`to_json` in the same namespace as the json types can fall afoul of "Unused function from_json" warnings/errors when compiling with flags `-Wunused-function` or `-Wall`.

This change makes the definitions `static inline` to avoid such errors as it is quite common if e.g. logging to only ever use the `to_json` methods.

### DIFF
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
@@ -15,7 +15,7 @@
 #include <unordered_set>
 #include <vector>
 
-static void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
+static inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
     if (j.contains("booleanData")) {
         j.at("booleanData").get_to(result.booleanData);
     }
@@ -65,7 +65,7 @@ static void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
         j.at("myFlags").get_to(result.myFlags);
     }
 }
-static void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
+static inline void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
     j = nlohmann::json {
         {"booleanData", item.booleanData},
         {"integer8Data", item.integer8Data},
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
@@ -65,7 +65,7 @@ inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
         j.at("myFlags").get_to(result.myFlags);
     }
 }
-static inline void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
+inline void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
     j = nlohmann::json {
         {"booleanData", item.booleanData},
         {"integer8Data", item.integer8Data},
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
@@ -15,7 +15,7 @@
 #include <unordered_set>
 #include <vector>
 
-static inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
+inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
     if (j.contains("booleanData")) {
         j.at("booleanData").get_to(result.booleanData);
     }
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
@@ -8,7 +8,7 @@
 #include <algorithm>
 #include <nlohmann/json.hpp>
 
-static inline void to_json(nlohmann::json& j, enum_data e)
+inline void to_json(nlohmann::json& j, enum_data e)
 {
     static const std::pair<enum_data, nlohmann::json> m[] = {{enum_data::FIRSTENUMVALUE,"FirstEnumValue"},{enum_data::SECONDENUMVALUE,"SecondEnumValue"}};
     auto it = std::find_if(std::begin(m), std::end(m),
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
@@ -8,7 +8,7 @@
 #include <algorithm>
 #include <nlohmann/json.hpp>
 
-static void to_json(nlohmann::json& j, enum_data e)
+static inline void to_json(nlohmann::json& j, enum_data e)
 {
     static const std::pair<enum_data, nlohmann::json> m[] = {{enum_data::FIRSTENUMVALUE,"FirstEnumValue"},{enum_data::SECONDENUMVALUE,"SecondEnumValue"}};
     auto it = std::find_if(std::begin(m), std::end(m),
@@ -18,7 +18,7 @@ static void to_json(nlohmann::json& j, enum_data e)
     });
     j = ((it != std::end(m)) ? it : std::begin(m))->second;
 }
-static void from_json(const nlohmann::json& j, enum_data& e)
+static inline void from_json(const nlohmann::json& j, enum_data& e)
 {
     static const std::pair<enum_data, nlohmann::json> m[] = {{enum_data::FIRSTENUMVALUE,"FirstEnumValue"},{enum_data::SECONDENUMVALUE,"SecondEnumValue"}};
     auto it = std::find_if(std::begin(m), std::end(m),
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
@@ -18,7 +18,7 @@ static inline void to_json(nlohmann::json& j, enum_data e)
     });
     j = ((it != std::end(m)) ? it : std::begin(m))->second;
 }
-static inline void from_json(const nlohmann::json& j, enum_data& e)
+inline void from_json(const nlohmann::json& j, enum_data& e)
 {
     static const std::pair<enum_data, nlohmann::json> m[] = {{enum_data::FIRSTENUMVALUE,"FirstEnumValue"},{enum_data::SECONDENUMVALUE,"SecondEnumValue"}};
     auto it = std::find_if(std::begin(m), std::end(m),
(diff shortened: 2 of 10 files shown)


src/main/scala/djinni/CppGenerator.scala
```diff
@@ -217,37 +217,38 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
                   )
                   .mkString(",")
 
-            w.wl(s"static void to_json(nlohmann::json& j, ${ident.name} e)")
```
mutagene, line 220: :eyes: scalafmt - nothing was changed other than the addition of the `inline` specifier.

jothepro:
I understand that the warnings by the compiler are not necessarily helpful in this case since the functions are generated and cannot be eliminated individually just because some of theme are not used.

But I have a few questions to help me better understand the provided solution:
- Did you test if this works with all major compilers? It seems like this may only hide the error in gcc but not in clang?
  > CC explicitly does not warn for unused static inline functions for 
-Wunused-function. [...]
  > Clang does warn for static inline functions that are unused.  
  
  ([Source](https://groups.google.com/g/linux.kernel/c/P9brByBnayI))
- From my understanding the compiler is not wrong when he is complaining that the function is unused. Did you consider to explicitly disable the warning with the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute instead of implicitly hiding it with the `inline` keyword?
- I am trying to imagine if there is a situation where one is actually interested in these warnings and it would be a problem if they don't occur. Do you think that's a valid scenario as well?

Bear with me if my questions indicate that I may not have fully grasped the issue yet. The `inline` keyword is new to me and I had to do some research first, which lead me to the above questions.

a4z:
This is a generated header, from some point of view like a system header.
As such, it should not cause any warnings in user code.

This is usually archived by including such headers from a directory specified as `-isystem`. Then there will be no warnings.
But the user needs to take care of that, specifying `-isystem` correctly.

From that point of view, we have nothing to do.

On the other hand, I think having an `inline` on functions that are implemented in a header that is not a template is correct. Otherwise, we could end up with multiple implementations of the same function if the header is included more than once. And therefore should be done. But I do not think we need the static keyword in this context.

On the other hand, for the template classes, we should not need the `inline` since templates should be inline per definition.
 
So yes, I think some things can be done with that code to make it look better, and this PR works in the right direction.

Can we give it a try, inline for free functions and no inline for templates?
And if there are warnings, our tests should cover them ... in an optimal world.
This needs of course not be done in the context of this PR, because I do not want to add a lot of work to @mutagene . So maybe a follow-up ticket could be nice ...

a4z:
Hi @mutagene , I would like to get that change in
But do you have the possibility to check if that would work, 
no static and just inline for the free functions and no inline for template functions?

Would be awesome if you could, since you also seem to have code that triggers problems, in case there are some.

mutagene:
Sorry for the delay. I'm really grateful for your thoughtful feedback and I'm sorry I haven't had time to respond properly. I'll take a look soon.

a4z:
Hi @mutagene , we recently added some fix to the generator. That speaks for a new release.

Do you think this PR has a chance to receive some updates the next time? 
Because if it does, we could well wait to have this improvement onboard.

src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
```diff
@@ -15,7 +15,7 @@
 #include <unordered_set>
 #include <vector>
 
-static void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
+static inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
```
a4z, line 18:
```suggestion
inline void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
```

src/it/resources/expected/all_datatypes_json/cpp-headers/all_datatypes_json+json.hpp
```diff
@@ -65,7 +65,7 @@ static void from_json(const nlohmann::json & j, all_datatypes_json& result)  {
         j.at("myFlags").get_to(result.myFlags);
     }
 }
-static void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
+static inline void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
```
a4z, line 68:
```suggestion
inline void to_json(nlohmann::json & j, const all_datatypes_json & item)  {
```

src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
```diff
@@ -8,7 +8,7 @@
 #include <algorithm>
 #include <nlohmann/json.hpp>
 
-static void to_json(nlohmann::json& j, enum_data e)
+static inline void to_json(nlohmann::json& j, enum_data e)
```
a4z, line 11:
```suggestion
inline void to_json(nlohmann::json& j, enum_data e)
```

src/it/resources/expected/all_datatypes_json/cpp-headers/enum_data+json.hpp
```diff
@@ -18,7 +18,7 @@ static void to_json(nlohmann::json& j, enum_data e)
     });
     j = ((it != std::end(m)) ? it : std::begin(m))->second;
 }
-static void from_json(const nlohmann::json& j, enum_data& e)
+static inline void from_json(const nlohmann::json& j, enum_data& e)
```
a4z, line 21:
```suggestion
inline void from_json(const nlohmann::json& j, enum_data& e)
```

src/it/resources/expected/all_datatypes_json/cpp-headers/my_flags+json.hpp
```diff
@@ -8,7 +8,7 @@
 #include <algorithm>
 #include <nlohmann/json.hpp>
 
-static void to_json(nlohmann::json& j, my_flags e)
+static inline void to_json(nlohmann::json& j, my_flags e)
```
a4z, line 11:
```suggestion
inline void to_json(nlohmann::json& j, my_flags e)
```

src/it/resources/expected/all_datatypes_json/cpp-headers/my_flags+json.hpp
```diff
@@ -20,7 +20,7 @@ static void to_json(nlohmann::json& j, my_flags e)
         }
     }
 }
-static void from_json(const nlohmann::json& j, my_flags& e)
+static inline void from_json(const nlohmann::json& j, my_flags& e)
```
a4z, line 23:
```suggestion
inline void from_json(const nlohmann::json& j, my_flags& e)
```

src/it/resources/expected/all_json_specialized_datatypes_in_custom_namespace/cpp-headers/my_enum+json.hpp
```diff
@@ -20,7 +20,7 @@ static void to_json(nlohmann::json& j, my_enum e)
     });
     j = ((it != std::end(m)) ? it : std::begin(m))->second;
 }
-static void from_json(const nlohmann::json& j, my_enum& e)
+static inline void from_json(const nlohmann::json& j, my_enum& e)
```
a4z, line 23:
```suggestion
inline void from_json(const nlohmann::json& j, my_enum& e)
```

src/it/resources/expected/all_json_specialized_datatypes_in_custom_namespace/cpp-headers/my_flags+json.hpp
```diff
@@ -10,7 +10,7 @@
 
 namespace custom_namespace {
 
-static void to_json(nlohmann::json& j, my_flags e)
+static inline void to_json(nlohmann::json& j, my_flags e)
```
a4z, line 13:
```suggestion
inline void to_json(nlohmann::json& j, my_flags e)
```

src/main/scala/djinni/CppGenerator.scala
```diff
@@ -217,37 +217,38 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
...
-                      w.wl("j.push_back(flagOption.second);")
-                    }
-                  }
-                } else {
-                  w.wl(
-                    s"static const std::pair<${ident.name}, nlohmann::json> m[] = {$enumInitializer};"
-                  )
-                  w.wl("auto it = std::find_if(std::begin(m), std::end(m),")
+            w.wl(
+              s"static inline void to_json(nlohmann::json& j, ${ident.name} e)"
```
a4z, line 221:
```suggestion
              s"inline void to_json(nlohmann::json& j, ${ident.name} e)"
```

src/main/scala/djinni/CppGenerator.scala
```diff
@@ -217,37 +217,38 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
...
+                  w.wl("return ej_pair.first == e;")
+                }
+                w.wl(
+                  "j = ((it != std::end(m)) ? it : std::begin(m))->second;"
+                )
               }
+            }
             w.wl(
-              s"static void from_json(const nlohmann::json& j, ${ident.name}& e)"
+              s"static inline void from_json(const nlohmann::json& j, ${ident.name}& e)"
```
a4z, line 251:
```suggestion
              s"inline void from_json(const nlohmann::json& j, ${ident.name}& e)"
```

src/main/scala/djinni/CppGenerator.scala
```diff
@@ -530,7 +531,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
             val recordSelf = ident.name
             // From JSON
             w.w(
-              s"static void from_json(const nlohmann::json & j, ${recordSelf}& result) "
+              s"static inline void from_json(const nlohmann::json & j, ${recordSelf}& result) "
```
a4z, line 534:
```suggestion
              s"inline void from_json(const nlohmann::json & j, ${recordSelf}& result) "
```

src/main/scala/djinni/CppGenerator.scala
```diff
@@ -557,7 +558,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
             }
             // To JSON
             w.w(
-              s"static void to_json(nlohmann::json & j, const $recordSelf & item) "
+              s"static inline void to_json(nlohmann::json & j, const $recordSelf & item) "
```
a4z, line 561:
```suggestion
              s"inline void to_json(nlohmann::json & j, const $recordSelf & item) "
```

a4z (approved):
Thank you!

I added some suggestions since I think the `static inline` is redundant, and the `static` can be removed.

Klick through them if you want or leave it; we will take that in the future. It's more of a cosmetic fix.

mutagene:
> Thank you!
> 
> I added some suggestions since I think the `static inline` is redundant, and the `static` can be removed.
> 
> Klick through them if you want or leave it; we will take that in the future. It's more of a cosmetic fix.

Yes, I completely agree.

a4z:
@jothepro , can you please glance over the PR and merge it if you think it's OK?
(I would appreciate a second pair of 👀 since I added changes myself)

jothepro:
I plan to do a full review tomorrow. 👍

jothepro (approved):

PR #143, opened 2022-10-27
[142] Shift flags into unsigned integers

### LINKED ISSUE #142: 32 bit flag to enum generation issue
In case of defining flags which later are translated to 32 bit unsigned enum, for the 32th enum value issue is occurred as '1<<31' is considered signed i32 operation and is overflow.
Instead, by generating with unsigned postfix 'u', i.e. '1u<<31' will work fine.

Proposal, use '1u<<shift' instead of '1 << shift'. Both C++ and ObjC support it fine

a4z:
Fixes #142

### DIFF
--- a/src/it/resources/expected/all_datatypes_json/cpp-headers/my_flags.hpp
+++ b/src/it/resources/expected/all_datatypes_json/cpp-headers/my_flags.hpp
@@ -9,9 +9,9 @@
 enum class MyFlags : unsigned {
     NO_FLAGS = 0,
     /** flag option comment */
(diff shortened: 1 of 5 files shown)


martin00001313 (approved):
that's the solution! thank you

jothepro (approved):

a4z:
We need to agree if this PR is a new release. 
I am OK with it but can not make it today; maybe tomorrow evening. I need to go to bed now; 
If anyone flags a new release, please go!

PR #155, opened 2023-10-30
Prevent methods renaming in Swift

niparx:
If I'm using generated code in Swift project, then Swift will rename my methods into something else (they have some naming rules). That, basically brakes all thing: API then looking totally different. There is a macro `NS_SWIFT_NAME`, where you can put a name, which Swift will use (something like "I know what I'm doing. Please, Swift, leave method name as it is written in macro."). So, there is an option for that: to add or not to add this macro to every method of generated API.

Some more context about this Swift feature:
- https://developer.apple.com/documentation/swift/renaming-objective-c-apis-for-swift
- https://github.com/apple/swift-evolution/blob/main/proposals/0005-objective-c-name-translation.md
- https://github.com/apple/swift/blob/main/docs/CToSwiftNameTranslation.md
- https://github.com/apple/swift/blob/main/docs/CToSwiftNameTranslation-OmitNeedlessWords.md

### DIFF
--- a/src/main/scala/djinni/Main.scala
+++ b/src/main/scala/djinni/Main.scala
@@ -87,6 +87,7 @@ object Main {
     var objcppIncludeObjcPrefixOptional: Option[String] = None
     var objcFileIdentStyleOptional: Option[IdentConverter] = None
     var objcStrictProtocol: Boolean = true
+    var objcPreventObjcFunctionRenameInSwift: Boolean = false
     var objcppNamespace: String = "djinni_generated"
     var cppCliOutFolder: Option[File] = None
     var cppCliIdentStyle = IdentStyle.csDefault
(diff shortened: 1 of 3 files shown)


a4z:
Thanks a lot for the PR!
It is excellent that you think about testing.

Tests are in the src/it folder.
For running them, please see [the running tests documentation](https://djinni.xlcpp.dev/djinni-generator/developer-guide/#running-tests)

Tests work basically so that there is an expected output file based on the given input. 

- `src/it/resources` has djinni files in the subfolder
-  `src/it/resources/expected` is the expected generated result from the djinni files based on the test
-  `src/it/resources/result` is created by the tests, 

The tests do compare the generated files with the expected ones.

The test calls are in `src/it/scala/djinni/GeneratorIntegrationTest.scala`
From there, you call djinni with arguments and then compare the result with the expected content.

a4z:
I haven't forgotten about the PR. I am just waiting on a little if tests will be added or not.
(The thumbs-up on my explanation/comment makes me kind of optimistic that we might get some)

niparx:
> I haven't forgotten about the PR. I am just waiting on a little if tests will be added or not. (The thumbs-up on my explanation/comment makes me kind of optimistic that we might get some)

Not forgotten. Writing tests + some minor improvements after feedback with Swift devs.

niparx:
Converting to `Draft`, since still `in progress`

PR #156, opened 2023-11-02
feat(deprecated): generate language specific deprecation markup

eakoli:
If @deprecated is present in a comment, a deprecation marker will be generated.

@deprecated is expected to be at the start of a line, any text following it on the same line is added as the ‘message’ text where supported

This applies to records/attributes/constants, interfaces/methods/constants, enums/values, flags/values

Supports C++, Java and ObjC generators


a4z:
Wow, this is nice! Thank you so much!!

a4z (approved):
Great PR! 
We need to add this to the documentation.
But that can happen once it's merged.

jothepro:
Could be nice if this would also support setting the `ObsoleteAttribute` in C++/CLI! @eakoli would you be interested in adding it in this PR or should I open a separate issue for it?

jothepro (approved):
apart from that it's looking great, thank you so much!

a4z:
If no other comments are made, I will soon merge this PR, and we will handle eventual ideas and additions, like docs, as extra issues.

a4z:
Thanks a lot!
This requires a new release, which will follow soon

PR #157, opened 2023-11-06
Import Web Assembly support from snapchat's djinni fork

mutagene:
I felt quite inspired by @eakoli 's latest contribution. I'm not sure there's the appetite for this, but at least on my projects, if I'm using djinni in the future it's hard to imagine not wanting the possibility of a web assembly (and flutter?) target. This is quite a naive attempt to steal that from the snapchat djinni, and without the lib-side work (which I haven't done yet), I'm not sure it even works. Anyway, the way these things go I may have time to work on it some more and I may not, so I thought I'd make the (draft) PR now and then worry about the rest later.


a4z:
Wow, that's a pretty fantastic development! 

Just a question: After looking back at the Python addition, I see that it is now 80% to 90% done but then slowed down.
Is there a danger we will end up in a similar situation (which might also be OK, because better 80% than nothing)
Or can we have a sample Hello World app utilizing this WASM generator once it is no longer in draft state?

mutagene:
Yes, I think we need some kind of definitive sample for this to be moved out of draft status. I haven't checked what there is in the snapchat repository, yet. I understand your  concerns about abandoned platforms, but (at least for me) web assembly is pretty essential. I think that, given the state of the support on the snapchat fork, it should be possible to get the web assembly support here quite usable without too much pain. Fingers crossed. Next up the support lib, I guess.

src/it/resources/expected/all_datatypes_json/cpp-headers/my_flags.hpp
```diff
@@ -9,10 +9,10 @@
 enum class MyFlags : unsigned {
     NO_FLAGS = 0,
     /** flag option comment */
-    FLAG1 = 1u << 0,
-    FLAG2 = 1u << 1,
-    FLAG3 = 1u << 2,
```
mutagene, line 14: @a4z I feel badly about this change especially in light of the change log for 1.4.0. If you can help me maintain the previous behaviour, I would be quite happy for that.

a4z: I can give it a look, but have to warn you that the next few days are pretty busy for me so it might take a while. 
Sorry to not be able to support your velocity at the moment.

src/main/scala/djinni/generator.scala
```diff
@@ -716,26 +759,27 @@ abstract class Generator(spec: Spec) {
...
-      ident: IdentConverter
+      ident: IdentConverter,
+      delim: String = "="
   ): Unit = {
     var shift = 0
     for (o <- normalEnumOptions(e)) {
       writeDoc(w, o.doc)
       w.wl(
-        ident(o.ident.name) + (if (e.flags) s" = 1u << $shift" else "") + ","
+        ident(o.ident.name) + (if (e.flags) s" $delim 1 << $shift"
```
a4z, line 781: needs a closer look, the `1` might / should be `1u`

src/main/scala/djinni/generator.scala
```diff
@@ -744,16 +788,17 @@ abstract class Generator(spec: Spec) {
...
       o <- e.options.find(_.specialFlag.contains(Enum.SpecialFlag.AllFlags))
     ) {
       writeDoc(w, o.doc)
-      w.w(ident(o.ident.name) + " = ")
+      w.w(ident(o.ident.name) + s" $delim ")
       w.w(
-        normalEnumOptions(e)
-          .map(o => ident(o.ident.name))
+        normalEnumOptions(e).zipWithIndex
+          .map { case (o, i) => s"(1 << $i)" }
```
a4z, line 801: needs a closer look, 
We write now default values to enums. This changes expected outcome, and the reason might be here

src/it/scala/djinni/GeneratorIntegrationTest.scala
```diff
@@ -967,4 +970,35 @@ class GeneratorIntegrationTest extends IntegrationTest with GivenWhenThen {
...
+      cppOmitDefaultRecordCtor = true
+    )
+
+    djinni(cmd)
+
+    Then(
+      "the @deprecated comments are generated as @Deprecated annotations"
+    )
+    assertFileContentEquals(idlFile, TS, tsFilenames)
+  }
```
eakoli, line 1003: It doesn't look like deprecation support was added for wasm, so I don't think this test is needed.

Also Im not sure why the enums had to be changed to use explicit values.

mutagene: That's true that I did nothing special for deprecation support. As I understand, though, the @deprecated tag matches the JSDoc spec ( https://www.typescriptlang.org/play#example/jsdoc-deprecated ) so I figured it was worth locking this expectation in place for generated TS. 

Re: enums - I get the impression that this was done in snapchat-djinni because it was expedient to have identical code generating enums for all languages. And it was easier for me, at least initially, to make the same assumption. But since the branches are now building for the generator and the support lib, I'd be quite happy to try to revert the behaviour to how it was before.

src/main/scala/djinni/generator.scala
```diff
@@ -716,26 +759,28 @@ abstract class Generator(spec: Spec) {
...
+      ident: IdentConverter,
+      delim: String = "=",
+      optionSuffix: String = ""
   ): Unit = {
     var shift = 0
     for (o <- normalEnumOptions(e)) {
       writeDoc(w, o.doc)
       w.wl(
-        ident(o.ident.name) + (if (e.flags) s" = 1u << $shift" else "") + ","
+        ident(o.ident.name) + (if (e.flags) s" $delim 1$optionSuffix << $shift"
```
mutagene, line 782: A bit hideous, but... 🤷‍♂️

a4z: Do we want to be super configurable on that level? 
It adds complexity, and complexity kills projects. And confuses users.
I do not see any return value for adding this option, but I am open for arguments

If possible, and there is no solid reason for the change, like a bug, I would like to keep the expected files in testing as they are.

mutagene: Ahh, this option is to restore the former behaviour so that non-wasm test file expectations are unchanged. How best to achieve this - by overriding behaviour in generator subclasses, or by making this method take more arguments to be more configurable - it's not clear to me. I think I would do things very differently if I were starting from scratch, but yeah, this is not that.

eakoli: Does wasm not support the current generated values of `1u << 4` ?

a4z: Same question here, can't wasm have the same declaration with flag + u values and enums with no values as the pure C++ part?

mutagene: wasm is like :

```
    EM_JS(void, djinni_init__my_enum_consts, (), {
        Module.MyEnum =  {
            /** enum option comment */
            OPTION1 : 0,
            OPTION2 : 1,
            OPTION3 : 2,
        }
    })
```

mutagene: I should probably post this instead -

`
    EM_JS(void, djinni_init__my_flags_consts, (), {
        Module.MyFlags =  {
            NO_FLAGS : 0,
            /** flag option comment */
            FLAG1 : 1 << 0,
            FLAG2 : 1 << 1,
            FLAG3 : 1 << 2,
            ALL_FLAGS : 0 | (1 << 0) | (1 << 1) | (1 << 2),
        }
    })
`

So no unsigned suffix, and I haven't checked, but I assume the "ALL_FLAGS" option cannot take the OR of the other flags because they are not yet set.

a4z: That might mean we must figure out in what generation mode we are and select the generation accordingly.
 Or there is another way ... but that would probably mean more re-write.
That sound like the `$optionSuffix` is the way to go, or do I misunderstand?

eakoli: for EM_JS, the content in {} is evaluated as javascript, so anything you can do In JS is valid.

mutagene: I think the $optionSuffix approach is simplest, because it only involves editing `generator.scala`. Otherwise I think we'd have to make a common base class for the generators that have unsigned flags, or maybe a mixin? But in the end I think the $optionSuffix makes sense here.

eakoli: probably a simpler option would be to pass a closure to allow the caller to formulate the value as needed.

something like 
```
def writeEnumOptions( w: IndentWriter, e: Enum, ident: IdentConverter, flag: ( Enum.Option, Int ) => String ) {
  var shift = 0
     for (o <- normalEnumOptions(e)) {
       writeDoc(w, o.doc)
       w.wl(
         ident(o.ident.name) + (if (e.flags) flag( o, shift ) else "") + ","
       )
       shift += 1
}  
```

then the default can be 
```
def writeEnumOptions( w: IndentWriter, e: Enum, ident: IdentConverter, flag: ( Enum.Option, Int ) => String ) {
    writeEnumOptions( w, e, ident, ( o: Enum.Option, shift: Int) => s" = 1u << $shift" );
}
```

and any generator can choose use a different flag value generation

eakoli: > Otherwise I think we'd have to make a common base class for the generators

Any generator can provide/use an alternate routine for generating its content it doesn't have to use the common routines in generator.

This might be preferable as to not complicate the common case for a single outlier that needs a specific alternate syntax.

Which I think hands down would be the simpler approach.

eakoli:
One of my general concerns with this PR is how many changes seem be outside of the new generator code, as in theory, adding a new generator should change little of the existing code other than extending the places to enable the new output.

E.g. Ive started working on a swift and kotlin generator and so far haven't had to make any changes to existing code other than adding the options and extra bools to enable them during testing

src/main/scala/djinni/generator.scala
eakoli: It looks like there would also need be changes to the writeEnumOptionsNone

a4z:
I agree, we need to check that all changes are related to the WASM part and not sneak in via any other Snapchat additions that might change behavior.

On the other hand, if the expected files from the tests are unchanged, nothing should have changed (or our tests are incomplete). So we should have some safety net in place,

mutagene:
> One of my general concerns with this PR is how many changes seem be outside of the new generator code, as in theory, adding a new generator should change little of the existing code other than extending the places to enable the new output.
> 
> E.g. Ive started working on a swift and kotlin generator and so far haven't had to make any changes to existing code other than adding the options and extra bools to enable them during testing

I've been trying to reduce these changes. I think all that remains is the `isOptional` change/fix, which I think/hope is tolerable given that it has not forced any changes in existing expected generated c++ code.

a4z:
I am sorry, but looking at my calendar, it is very likely my review must wait until the Christmas holidays.
I want to prioritize these tasks, and I will try to do it, but it is difficult. 
Thank you for your understanding and patience.

a4z:
After thinking about this PR for one year, I tend to not use it.
We had multiple language extensions dropped into djinni, C, C#, and Python, which are no longer supported by the original authors but are creating a huge maintenance burden. 
This is why I am hesitant about new language extensions, 

In hindsight, it would have been beneficial to keep the original two language bindings, Objective-C and Java/JNI, and focus only on those. And add anything else in a djinni2, based on different ideas and technologies. 

That is probably what should happen for this project.

mutagene:
> After thinking about this PR for one year, I tend to not use it. We had multiple language extensions dropped into djinni, C, C#, and Python, which are no longer supported by the original authors but are creating a huge maintenance burden. This is why I am hesitant about new language extensions,
> 
> In hindsight, it would have been beneficial to keep the original two language bindings, Objective-C and Java/JNI, and focus only on those. And add anything else in a djinni2, based on different ideas and technologies.
> 
> That is probably what should happen for this project.

I can completely understand that. I think that PyDjinni has done a nice job of making language bindings a bit more modular, but with the original Djinni it is difficult to do so. 

While I like quite a bit how this project is organised and operated, it's a pity to me that it will continue to diverge from Snap's fork and not support web. If one could only choose two platform languages, then I agree those two should be for Java and Android. For a third platform, personally I think web/webassembly makes the most sense - but in any case, ifi need web, I can just use Snap's fork. By all means, feel free to close the PR.

a4z:
The question is, @mutagene , would you be willing to join as maintainer and help maintain the Webassembly implementation? And commit to that.
If you do, we probably can merge that branch.

a4z:
Since I plan to step back to the original djinni, Java Objective-C , to keep this repo maintainable, no new languages will land here. If that is desired, we need a djinni2 that is tailord to 2026 and beyond

PR #159, opened 2023-11-11
Maintenance toolchain update 2023

a4z:
Since we are going to have a release soon, 
It would be a good idea to update scala and sbt also
and also check the current dependencies if they need any update


build.sbt
```diff
@@ -1,11 +1,16 @@
 import sbtassembly.AssemblyPlugin.defaultUniversalScript
 
-ThisBuild / scalaVersion := "2.13.6"
+ThisBuild / scalaVersion := "2.12.18"
```
a4z, line 3: This is intentional, 2.13 brings 100 + deprecations, 
so first, get a clean 2.12 state, and then move on

a4z:
This is quite a huge update, but it makes scala 2.12.18 (latest 2.12) work with all deprecation warnings and other warnings fixed. 
(2 unused variable warnings are left, but I think this is an error in the tool) 

2.13 will bring a huge bulk of new deprecation warnings, so I keep this for a later update, since 2.12 is still maintained.

mutagene:
Don't suppose you can wait until I finish with the web assembly support🤞

a4z:
It depends on the timing. I would like to have this out and make a release with the new functionality.
Also, pin the Scala version and add the tools. This will help with newly written code.

This might cause merge issues. No worries, I can help resolve them.
I will not leave you alone in case of problems.  The plan is helping you to merge main into your branch, and have all the new checks applied to the new code.

Btw, this PR shows quite well the maintenance cost of code that is not in use, like the Python generator. And why I ask for additions of code that has more to show than some unit test, like a small demo app

mutagene:
It's fine.  Worried about conflicts, but really any conflicts from this PR are nothing next to what has to be done with the support lib 😅

a4z:
Don't plan to add any more changes for now. 

Hope I will get at least some review, to not be forced to bypass our review policies

freitass (approved):
Sorry for the delay, and thanks for keeping this project alive :)

jothepro (approved):

PR #177, opened 2025-09-24
Remove Python and C++/CLI language bindings

a4z:
- Removed all Python generators, marshals, and related code
- Removed all C++/CLI generators, marshals, and related code
- Removed all Python and C++/CLI test files and expected outputs
- Updated CLI to remove --py-out and --cppcli-out arguments
- Updated documentation to reflect removed language support
- All integration tests now pass (27/27)
- Core functionality for remaining languages (Java, C++, Obj-C, Obj-C++) preserved

PR #179, opened 2025-09-24
Build/bzl

a4z:


a4z:
requires #177 to be merged first

